PR #45, opened 2019-04-09
39 shellcheck

mcurran-inviqa:


.travis.yml
```diff
@@ -33,6 +33,7 @@ script:
   - ws exec composer test-quality
   - ws exec composer test-unit
   - ws exec composer test-acceptance
+  - find . -type f ! -path "./.git/*" ! -path "./bin/workspace" ! -path "./bin/build" ! -path "./vendor/*" ! -name "*.phar" ! -name "*.php" -and \( -name "*.sh" -or -perm -0111 \) -exec shellcheck --exclude=SC1008,SC1091 {} +
```
hgajjar, line 36: I think the paths are incorrect, it should be something like `find . -type f ! -path "./.git/*" ! -path "./dist/*" ! -path "./test/*" ! -name "*.xml" ! -name "*.txt" ! -name "*.properties" -and \( -name "*.sh" -or -perm -0111 \) -exec shellcheck --exclude=SC1008,SC1091 {} +`

hgajjar (commented):
You also need to fix the shellcheck errors. I was also working on this issue and already have a working branch ready, I can push it if you don't mind?

.travis.yml
kierenevans: I'd argue the solr config should not be executable, but we can deal with that in another PR.

mcurran-inviqa:
@hgajjar will leave it with you and close this pr, can you assign the issue to yourself if I see an issue is not assigned I will assume ok to pickup

PR #46, opened 2019-04-09
32 behat chrome

mcurran-inviqa:


src/drupal8/application/skeleton/composer.json
```diff
@@ -13,18 +13,14 @@
...
-      "url": "https://github.com/my127/behat-chrome-extension"
     }
   },
   "require-dev": {
     "webflo/drupal-core-require-dev": "^8",
     "drupal/devel": "^2.0",
     "drupal/drupal-extension": "^3.4",
     "mikey179/vfsstream": "^1.0",
-    "my127/behat-chrome-extension": "^1.2"
+    "dmore/behat-chrome-extension": "^1.3"
```
kierenevans: Please reorder to be alphabetical (same for all)

kierenevans (changes requested):

src/drupal8/application/skeleton/composer.json
tkotosz: maybe use
```
"config": {
        "sort-packages": true
    },
```
https://getcomposer.org/doc/06-config.md#sort-packages
?

tkotosz: ah it is already used :D I guess the composer.json was edited manually :D

kierenevans (approved):
Thanks!

PR #47, opened 2019-04-09
Add shellcheck to fix common errors

### LINKED ISSUE #39: Add shellcheck to catch common errors
Configure travis to run shellcheck to catch common issues like the one highlighted in https://github.com/inviqa/harness-base-php/issues/36

Example can be seen at :- https://github.com/my127/workspace/blob/0.1.x/.travis.yml#L17

**AC**
- [x] shellcheck added to .travis.yml 
- [x] errors picked up by shellcechk are resolved

hgajjar:
Fixes #39

### DIFF
--- a/src/_base/harness/scripts/destroy.sh
+++ b/src/_base/harness/scripts/destroy.sh
@@ -9,6 +9,7 @@ if [[ "$USE_DOCKER_SYNC" = "yes" ]]; then
 fi
 
 if [[ "$APP_BUILD" = "static" ]]; then
+    # shellcheck disable=SC2046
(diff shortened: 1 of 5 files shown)


src/_base/harness/scripts/destroy.sh
```diff
@@ -9,6 +9,7 @@ if [[ "$USE_DOCKER_SYNC" = "yes" ]]; then
 fi
 
 if [[ "$APP_BUILD" = "static" ]]; then
+    # shellcheck disable=SC2046
```
hgajjar, line 12: As we are intentionally splitting the word here, I'm not sure if we can fix this shellcheck error.

coledave (approved):

PR #51, opened 2019-04-09
48 nginx vars

mcurran-inviqa:


kierenevans:
We should have an empty `nginx.conf` in the default attributes.

kierenevans (changes requested):
Thanks! Can we have an empty `nginx.conf` in the default attributes?

kierenevans (approved):

PR #52, opened 2019-04-09
11 allow specify m1 admin

mcurran-inviqa:


src/magento1/application/overlay/public/app/etc/local.xml.twig
```diff
@@ -77,7 +77,7 @@
         <routers>
             <adminhtml>
                 <args>
-                    <frontName><![CDATA[admin]]></frontName>
+                    <frontName><![CDATA[{{ @('admin.path') ?: 'admin' }}]]></frontName>
```
kierenevans: We should probably add this to magento1's harness.yml so that it's easily visible that it can be modified. Perhaps under the `magento.admin.path` scope?

mcurran-inviqa: 👍

mcurran-inviqa:
PR also covers https://github.com/inviqa/harness-base-php/issues/9 as it is so small and related

kierenevans (approved):

PR #56, opened 2019-04-09
28 wp tests

mcurran-inviqa:


mcurran-inviqa:
removed code as it does not do anything or conform to wp standards,

not sure we do want to conform but this pr adds them.

coledave (approved):

PR #58, opened 2019-04-10
10 mage run code

mcurran-inviqa:


kierenevans (changes requested):
Thanks! Can we have an empty `magento.run.code` and `magento.run.type` in the default attributes for the M1 harness?

kierenevans (approved):

PR #63, opened 2019-04-12
Spryker harness

### LINKED ISSUE #17: Create an environment for Spryker

hgajjar:
Fixes #17 

Genaral Questions:
- Single or individual services for yves/zed containers? Currently I have kept only one service which serves as both yves/zed, however splitting it would increase the number of services even more (there are 10 containers already), as it would require two groups of nginx/php-fpm services for yves and zed separately.
- Jenkins service - I have created two services for Jenkins, one for jenkins itself using official jenkins image and a jenkins slave which is created from console image, which connects to Jenkins master during initialization. Is this a good approach?


src/spryker/harness/scripts/enable.sh
```diff
@@ -0,0 +1,54 @@
...
+
+    if [[ "$USE_DOCKER_SYNC" = "yes" ]]; then
+        passthru docker-sync start
+        passthru docker-sync stop
+    fi
+
+    # this is causing issue with setting a local image
+    # passthru docker-compose pull
+    # passthru docker-compose build --pull
+    passthru docker-compose build
```
hgajjar: I've tag console image as ":dev" in local and using it as base image for jenkins-slave, because of this it gives an error when it tries to pull the images first (as local image for console doesn't exist yet), so I had to comment pull command. Is it ok to skip pull as build would do it anyway if image doesn't exist? or how should we handle this?

richardmiller-zz:
With regards to making Yves and Zed separate services I don't think it's essential and they are served from the same nginx instance on the Spryker VM. However, we have typically run them as separate services in production so it would make sense for them to be separate services at least for CI and preview environments (which I guess means they would be for dev as well). 

We have had issues with the connections between them in the past that have caused issues and having them as separate services may help with catching this earlier.

src/spryker/harness.yml
```diff
@@ -0,0 +1,78 @@
...
+      - chrome
+      - postgres
+      - redis
+      - elasticsearch
+      - rabbitmq
+      - jenkins
+      - jenkins-slave
+  spryker:
+    salt: 2tkqCCRKy5rT4wMVz8KTCh8r3sJGkL5v
+    demoshop-url: https://github.com/spryker-shop/b2c-demo-shop.git
```
richardmiller-zz, line 22: Is it possible to override this with the b2b demoshop if needed at the time of creating a project?

richardmiller-zz:
I don't think I have any real insight into the Jenkins questions - I'm not sure I understand the reasons for needing both services but that's my lack of experience with it not me questioning the approach.

src/spryker/harness.yml
hgajjar: That is my plan, not sure yet how to do that but keeping this value configurable in yml is the first step

hgajjar:
@richardmiller regarding Jenkins service, spryker requires the entire codebase and all php services/configurations to run a jenkins job. So,
- either we need to install Jenkins on top of "spryker console" image, which is not ideal as it would make the final docker image very big and it would be against the docker methodology (micro-services with single responsibility)
- or we can keep jenkins in a separate container and only add java run time on top of "spryker console" image, which allows it to connect to jenkins master and run as a slave, which is what I have done.

hgajjar:
> With regards to making Yves and Zed separate services I don't think it's essential and they are served from the same nginx instance on the Spryker VM. However, we have typically run them as separate services in production so it would make sense for them to be separate services at least for CI and preview environments (which I guess means they would be for dev as well).
> 
> We have had issues with the connections between them in the past that have caused issues and having them as separate services may help with catching this earlier.

I agree that in production we should have different services for yves/zed, but in local env it would make it really hard for a dev to figure out the location of a problem if something goes wrong in yves->zed communication. Though keeping it different in local/non-local envs makes it even harder to debug issues and it also adds an overhead of managing different configurations.

Any suggestions @dcole-inviqa @kierenevans ?

richardmiller-zz:
@hgajjar Thank you for the explanation of the Jenkins issue. I see why it makes sense to have the two services/images for this.

richardmiller-zz:
@hgajjar for the yves/zed question - it looks like the local docker set up for a recent spryker project uses a single service for both.

.travis.yml
```diff
@@ -19,6 +19,9 @@ env:
   # wordpress
   - HARNESS=wordpress MODE=static
   - HARNESS=wordpress MODE=dynamic
+  # spryker
```
kierenevans, line 19: Okay to move this to be alphabetical order?

build
```diff
@@ -3,7 +3,7 @@
 set -e
 set -x
 
-HARNESSES=(drupal8 magento1 magento2 wordpress)
+HARNESSES=(drupal8 magento1 magento2 wordpress spryker)
```
kierenevans: Okay to move this to be alphabetical order?

src/_base/_twig/docker-compose.yml/service/postgres.yml.twig
```diff
@@ -0,0 +1,12 @@
+  postgres:
+    image: postgres:9.4
```
kierenevans: 9.4 has a scheduled final release on February 13, 2020 according to https://www.postgresql.org/support/versioning/ 
I'm not sure from that page if it will be supported after. Is it possible to use a newer postgres version with spryker?

src/spryker/_twig/docker-compose.yml/environment.yml.twig
```diff
@@ -0,0 +1,24 @@
+      SPRYKER_SALT: {{ @('spryker.salt') }}
+      HOST_OS_FAMILY: {{ @('host.os') }}
+      APP_NAME: {{ @('workspace.name') }}
+      YVES_HOST: {{ @('yves-host') }}
```
kierenevans: The convention seems to be underscores up until now (e.g. `database.root_pass`), perhaps we should use underscores too, or even better, something like `yves.external_host` and `zed.external_host` (or `.public_host`)?

src/spryker/_twig/docker-compose.yml/service/elasticsearch.yml.twig
```diff
@@ -0,0 +1,8 @@
+  elasticsearch:
+    build: .my127ws/docker/image/elasticsearch
```
kierenevans: If there's a build, we will want a twig if/else for dynamic/static where static uses a pre-built docker image stored in the docker.repository (like you have with jenkins-runner)

src/spryker/_twig/docker-compose.yml/service/jenkins-slave.yml.twig
```diff
@@ -0,0 +1,23 @@
+  jenkins-slave:
```
kierenevans: While I know this is the terminology that Jenkins uses, perhaps `jenkins-runner` might be better.

src/spryker/harness.yml
```diff
@@ -0,0 +1,78 @@
...
+    host: rabbitmq
+    port: 5672
+    api_port: 15672
+    user: spryker
+    password: spryker
+    vhost: /spryker
+  jenkins:
+    host: jenkins
+    port: 8080
+  yves-host: = 'yves.' ~ @('namespace') ~ '.' ~ @('domain')
```
kierenevans: I recommend a hostname like: `yves-<namespace>.<domain>` due to the SSL certificate bundled with workspace only being valid for one level of wildcard.
Ditto for zed, jenkins and rabbitmq :)

src/spryker/_twig/docker-compose.yml/service/rabbitmq.yml.twig
```diff
@@ -0,0 +1,16 @@
+  rabbitmq:
+    image: rabbitmq:3.7-management
```
kierenevans: How about the alpine version - `rabbitmq:3.7-management-alpine`?

src/spryker/docker/image/console/root/lib/task/skeleton/apply.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+function task_skeleton_apply()
+{
+    if [ ! -f /app/composer.json ]; then
+      fetch_demoshop
```
kierenevans: Mixed indentation in this file

src/spryker/docker/image/elasticsearch/Dockerfile
```diff
@@ -0,0 +1,5 @@
+
+# https://github.com/elastic/elasticsearch-docker
+FROM docker.elastic.co/elasticsearch/elasticsearch:5.6.14
```
kierenevans: Elasticsearch 5.6 is EOL as of 2019-03-11 according to https://www.elastic.co/support/eol - does a newer version work with Spryker?

It looks like the ES images have migrated back to https://hub.docker.com/_/elasticsearch thankfully, so could use that source (and the `5.6` tag instead of an older version of `5.6` release to pick up fixes)

src/spryker/docker/image/jenkins-slave/root/lib/task/jenkins/setup.sh
```diff
@@ -0,0 +1,55 @@
...
+        echo -e "Already registered as jenkins slave $JENKINS_SLAVE_NAME:"
+        cat /tmp/jenkins.node
+    fi
+
+    # link php where it is expected by cron.conf
+    ln -s /usr/local/bin/php /usr/bin/php
+}
+
+function wait_for_jenkins_master() {
+    local counter=0
```
kierenevans: can be removed

src/spryker/docker/image/nginx/root/etc/nginx/conf.d/default.conf.twig
```diff
@@ -0,0 +1,96 @@
...
+
+        # Terminate OPTIONS requests immediately. No need for calling php
+        # OPTIONS is used by Ajax from http to https as a pre-flight-request
+        # see http://en.wikipedia.org/wiki/Cross-origin_resource_sharing
+        if ($request_method = OPTIONS) {
+            return 200;
+        }
+
+        add_header X-Server $hostname;
+        fastcgi_pass php-fpm:9000;
```
kierenevans, line 51: This passes every request apart from the static assets listed above to PHP, are there any access routes that should be denied or not allow PHP processing (e.g. upload directories)?

src/spryker/docker/image/nginx/root/etc/nginx/conf.d/default.conf.twig
```diff
@@ -0,0 +1,96 @@
...
+        expires           30d;
+        add_header Pragma public;
+        add_header Cache-Control "public, must-revalidate, proxy-revalidate";
+        try_files $uri =404;
+    }
+
+    # PHP application gets all other requests
+    location / {
+        add_header X-Server $hostname;
+        fastcgi_pass php-fpm:9000;
```
kierenevans, line 95: This passes every request apart from the static assets listed above to PHP, are there any access routes that should be denied or not allow PHP processing (e.g. upload directories)?

src/spryker/application/overlay/config/install/docker.yml
```diff
@@ -0,0 +1,255 @@
+# This file is a copy of https://github.com/spryker-shop/b2c-demo-shop/blob/master/config/install/development.yml
+# We have remove US store from database related steps as original file expects a separate database for US store
+# which is not the case in this setup
+
+env:
+    APPLICATION_ENV: development
```
kierenevans: Presuming this will need to change in the dynamic environments when we get there - a note for later.

src/spryker/harness/scripts/enable.sh
kierenevans: It would be better to still have the pull to avoid stale images causing inconsistent builds between people.

Perhaps for the first pull:
```bash
docker-compose config --services | grep -v console | xargs docker-compose pull
```

and for the build --pull:
```bash
docker-compose config --services | grep -v jenkins-runner | xargs docker-compose build --pull
docker-compose build jenkins-runner
```

src/spryker/_twig/docker-compose.yml/application.yml.twig
```diff
@@ -0,0 +1,63 @@
...
+      - traefik.backend={{ @('workspace.name') }}
+      - traefik.frontend.rule=Host:{{ @('yves-host') }},{{ @('zed-host') }}
+      - traefik.docker.network=my127ws
+      - traefik.port=80
+    links:
+      - php-fpm:php-fpm
+    networks:
+      private:
+        aliases:
+          - {{ @('hostname') }}
```
kierenevans, line 46: Should add the yves and zed external hosts here too so that behat can talk to them via the internal network.

kierenevans:
In response to the general questions:
1. It might be possible to keep 1 nginx container like currently but have 2 php-fpm containers to handle the different applications, though I guess in a load testing scenario that nginx instance will get blocked waiting for zed requests to complete when yves requests are trying to go through.
2. The current approach works for me regarding jenkins + it's runner.

kierenevans:
For the static CI build (in Jenkins), we'll need to override the `app build` command to build the extra docker images. (Base version is https://github.com/inviqa/harness-base-php/blob/0.2.x/src/_base/harness/config/build.yml )

src/spryker/docker/image/nginx/root/etc/nginx/conf.d/default.conf.twig
hgajjar: There is an `assets` directory, but that is handled here https://github.com/inviqa/harness-base-php/blob/aafebc1880e0dc1f6bfb68407a29f50c6b1e847c/src/spryker/docker/image/nginx/root/etc/nginx/conf.d/default.conf.twig#L14

src/spryker/docker/image/nginx/root/etc/nginx/conf.d/default.conf.twig
hgajjar: Same as above

src/spryker/application/overlay/config/install/docker.yml
hgajjar: I will move it to a variable for now, so it can be changed easily later on

hgajjar:
@kierenevans I have done almost all changes you requested, also fixed some bugs I found. Could you please review it again? thanks!

kierenevans (dismissed):
Absolutely amazing :star2:

src/_base/docker/image/console/root/lib/task/assets/apply.sh.twig
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+function task_assets_apply()
+{
+    local ASSETS_DIR="${ASSETS_DIR:-tools/assets/development}"
+    local IMPORT_COMMAND=""
+
+    if [ "{{ @('database.type')|raw }}" == "mysql" ]; then
```
andytson-inviqa, line 8: Just a thought here, rather than rendering dead code I think this would be better as either:
a) an environment variable - symfony even requires the variable for it's DBAL platform, so would kill two birds with one stone, tho has additionally semantic meaning for adding version on as well
b) use twig conditions to render only the relevant task output

I prefer (a) really

hgajjar: True, could you please open an issue for it?

hgajjar: ah, you already opened PR, even better 👍

PR #65, opened 2019-04-17
Remove settings.php from overlay and add to skeleton, settings.local.…

sgrichards:
* Remove settings.php from overlay and add to skeleton (settings.php must be committed & version-able by the project team) 
* `settings.local.php` added to overlay to support harness specific configuration 
(not sure if this is the correct place for it??)

_note: settings.local.php is not to be committed by the project team and is for environment specific (development) purposes._


coledave:
Thanks for looking into this, overlay is the correct place :) guessing we will want to add it to the gitignore in place of the settings.php

paulmartin84:
I believe settings.local.php is already ignored in https://github.com/inviqa/harness-base-php/blob/60324c6e72dbd1c159c733cbd68676d858008492/src/drupal8/application/skeleton/docroot/.gitignore#L22

kierenevans (approved):

PR #82, opened 2019-05-13
bash to stop at unhandled errors

marcomc:
added to every bash script
```bash
set -o errexit
set -o pipefail
set -o nounset
```


marcomc:
reverted changes: Will do better on another branch

PR #84, opened 2019-05-14
Issue-72: bash to stop at unhandled errors

marcomc:
add bash settings to control errors on all non-lib bash scripts

### DIFF
--- a/build
+++ b/build
@@ -1,6 +1,8 @@
 #!/bin/bash
 
-set -e
+set -o errexit
+set -o pipefail
+set -o nounset
 set -x
 
 HARNESSES=(drupal8 magento1 magento2 spryker wordpress)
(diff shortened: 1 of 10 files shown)


andytson-inviqa (changes requested):
There are unhandled errors that now need handling

Tips on ways to handle them:
```
if ! command-that-fails; then
  error handling here
fi
```
```
set +o errexit
some-lines that don't
need error handling
best to avoid this pattern tho
set -o errexit
```
```
set +o errexit
command-that-fails
local -r RETURN_CODE="$?"
set -o errexit

if [ "$RETURN_CODE" -gt 0 ]; then
    error handling here
fi
```
```
a-command-you-want-to-ignore-error-of || true
```

PR #94, opened 2019-05-22
Spryker mail configuration

hgajjar:


hgajjar:
Should we add this in base harness, so that other frameworks can use it too? `php-fpm` is linked to `nginx` in base harness already.

coledave (approved):

PR #95, opened 2019-05-22
Set default redis "save" configuration

hgajjar:
When we set any redis configuration via command argument, it resets all default redis configurations. However, only "save" configuration is being affected by this issue. It empties the value of "save", which prevents redis from persisting data on disk. More info: https://github.com/docker-library/redis/issues/87


kierenevans (approved):
Good find! Thanks

PR #103, opened 2019-05-24
Add setting toggle commands

kierenevans:
Convenience commands that insert into workspace.override.yml and apply the setting immediately:
* `ws xdebug on/off`
* `ws xdebug cli on/off`
* `ws docker-sync on/off`


hgajjar (approved):
LGTM

PR #106, opened 2019-05-27
Fix phpstan include with docker-sync

kierenevans:
https://github.com/inviqa/harness-base-php/blob/0.2.x/src/magento2/application/skeleton/phpstan.neon#L3 doesn't work with docker-sync due to us not syncing .my127ws to the docker-sync data container.

Allow this to be loaded by removing the ignore

### DIFF
--- a/src/_base/docker-sync.yml.twig
+++ b/src/_base/docker-sync.yml.twig
@@ -11,7 +11,6 @@ syncs:
     sync_excludes: &IGNORE
       - .docker-sync
       - .idea
-      - .my127ws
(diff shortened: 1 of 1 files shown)


hgajjar:
If we sync this directory then may be we don't need a separate volume for it (https://github.com/inviqa/harness-base-php/blob/0.2.x/src/_base/_twig/docker-compose.yml/application.yml.twig#L17), it was created for copying the php config files to enable disable xdebug.

hgajjar (approved):

PR #107, opened 2019-05-27
Spryker helm templates

hgajjar:
- Add spryker helm templates
- Allow setting global nginx configurations
- Allow using environment variables in nginx configurations, to let it use the same image for CI test build and preview environment build

### DIFF
--- a/src/_base/harness/config/pipeline.yml
+++ b/src/_base/harness/config/pipeline.yml
@@ -1,10 +1,9 @@
 command('app publish'): |
   #!bash(workspace:/)|@
-  run docker login -u="@('docker.username')" -p="@('docker.password')"
+  run docker login -u="@('docker.username')" -p="@('docker.password')" @('docker.repository')
   run docker push @('docker.repository'):@('app.version')-console
   run docker push @('docker.repository'):@('app.version')-php-fpm
   run docker push @('docker.repository'):@('app.version')-nginx
-  run docker push @('docker.repository'):@('app.version')-jenkins-runner
   run docker logout @('docker.repository')
 
 command('app deploy preview'): |
--- a/src/_base/harness/config/pipeline.yml
+++ b/src/_base/harness/config/pipeline.yml
@@ -1,9 +1,10 @@
 command('app publish'): |
   #!bash(workspace:/)|@
-  run docker login -u="@('docker.username')" -p="@('docker.password')" @('docker.repository')
+  run docker login -u="@('docker.username')" -p="@('docker.password')"
   run docker push @('docker.repository'):@('app.version')-console
   run docker push @('docker.repository'):@('app.version')-php-fpm
   run docker push @('docker.repository'):@('app.version')-nginx
+  run docker push @('docker.repository'):@('app.version')-jenkins-runner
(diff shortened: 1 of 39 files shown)


hgajjar:
@kierenevans @dcole-inviqa Regarding https://github.com/inviqa/harness-base-php/pull/107/commits/76fecb84c2137dc1e5e3c4432f7ec0d3f8b2216a, I'm preprocessing the nginx configurations to apply environment variables to it, I think this is a general problem to do templating before starting an image, should we have a global solution to this which can be used in other images too? (For ex, in CP dockerfiles, we had toml templating engine for this)

kierenevans:
@hgajjar I think the idea is to preprocess as much as possible using the twig templating. If it won't change after making the environment, no need for an environment variable.

hgajjar:
@dcole-inviqa @kierenevans How should we proceed here? David agreed with my approach of preprocessing nginx configurations, to see how it works in spryker harness and we can later move it to base harness too. There is also some work done to allow setting global nginx configurations, which needs some review too. Could you please review it again?

src/_base/harness/config/pipeline.yml
```diff
@@ -1,9 +1,10 @@
 command('app publish'): |
   #!bash(workspace:/)|@
-  run docker login -u="@('docker.username')" -p="@('docker.password')" @('docker.repository')
```
kierenevans: By removing the docker.repository, we can't log into for example `quay.io` or `gcr.io`, only docker hub.

src/spryker/helm/preview/templates/nginx-service.yaml
```diff
@@ -0,0 +1,20 @@
...
+metadata:
+  annotations:
+    kompose.cmd: kompose convert -c
+    kompose.version: 1.18.0 ()
+  creationTimestamp: null
+  labels:
+    io.kompose.service: nginx
+  name: nginx
+spec:
+  type: LoadBalancer
```
kierenevans: This will assign an IP address when on GKE/AWS. As you already have an ingress object, set this to ClusterIP or omit it to use the default.

kierenevans (changes requested):
Just two things :)

src/_base/harness/config/pipeline.yml
hgajjar: ok, so I should use repository as `docker.io/<repourl>` which also allows login to dockerhub.

kierenevans: No just need to put this back to how it was - set `docker.repository` in the projects as the whole domain/repo name, e.g. "hello/example" for docker hub (or "docker.io/hello/example"), or "quay.io/hello/example", etc.

hgajjar: "hello/example" works in docker pull but does not allow to login, only "docker.io/hello/example" works in both cases

kierenevans: True, perhaps we should split the domain from the repo name?

kierenevans (approved):

hgajjar:
Closing this PR to reuse the latest helm templates from base harness

PR #109, opened 2019-05-29
Reset console image to use build user by default

kierenevans:
Fixes direct docker usage without "-u build" creating files that UID 1000 can't modify in local development environments.


kierenevans:
Never mind, entrypoint needs root.

PR #120, opened 2019-06-18
Add migrate task to facilitate deployments

coledave:


src/drupal8/harness.yml
```diff
@@ -55,7 +55,7 @@ attributes:
         - run "chmod 755 /app/docroot/sites/default"
         - task "overlay:apply"
         - task "assets:dump"
-    init:
+    migrate:
```
andytson-inviqa, line 58: so, migrate doesn't run on local environments by default but init does, right? If so does this cause any issues?

PR #121, opened 2019-06-18
Add post-upgrade-hook to call the app migrate task

coledave:


kierenevans:
I think this should be a pre-upgrade hook to avoid downtime - it's easier to make the database forwards compatible with old and new code rather than the code.

kierenevans (approved):
Ace!

PR #129, opened 2019-06-26
Allow running commands at php-fpm entrypoint

hgajjar:


src/_base/helm/app/templates/application/php-fpm/deployment.yaml
```diff
@@ -31,6 +31,7 @@ spec:
             port: 9000
           initialDelaySeconds: 5
           periodSeconds: 10
+        command: ["/entrypoint.sh"]
```
kierenevans: Should we put this in the Dockerfile for php-fpm instead? Then only override it in the dynamic part in docker-compose.yml?

hgajjar: or just set both dynamic and static entrypoints in Dockerfile based on twig condition `{% if @('app.build') == 'static' %}` and remove it completely from docker-compose?

kierenevans: Yeah as it's now a template - good idea :)

kierenevans (dismissed):

hgajjar:
Hi @kierenevans , could you please rereview this PR?

kierenevans (dismissed):

PR #140, opened 2019-07-01
Add memory limits for k8s resources

hgajjar:


src/_base/harness/attributes/common.yml
```diff
@@ -135,6 +135,32 @@ attributes.default:
   helm:
     timeout: 300
 
+  resources:
+    requests:
+      memory:
+        console: "10Mi"
```
kierenevans, line 141: For cluster stability set these requests to the exact same as the limit. Perhaps even don't allow creating a difference?

hgajjar: discussed this with @dcole-inviqa :) 
It's true for other services, but for console, if we set it 2G, that would occupy it from beginning, which is a waste of memory as in general it would require very less, but if user wants to run some process with high memory then we should allow him to. 
So except "console", we have set it same for other services

kierenevans: This is still problematic though - let's say a node has 2058MiB (arbitrary example).

Kubernetes looks at the requests and schedules the following onto that node
* elasticsearch
* mysql
* postgres
* console

Elasticsearch, mysql and postgres use as much memory as given to them - 2048MiB.

Console is now left with 10MiB (what K8s has been told console needs for scheduling) so any higher memory command you not be able to run successfully due to OOMKill.

Depending on how the kernel handles the OOMKill, dockerd can even be killed and the node goes unresponsive.

hgajjar: true, I'm aware of this issue. The problem is, spryker's basic env would need more than 6GB memory if we also give 2G to console 😃

hgajjar: how about setting it same for requests/limit here in base harness, and change in project repo if still required? The idea of this PR is just to put some default limits and to enable projects setting limit without overwriting helm templates

hgajjar: ping @kierenevans

coledave (approved):

PR #142, opened 2019-08-02
Always pull when building images in CI

kierenevans:
Avoid out of date base images by pulling when building.

### DIFF
--- a/src/_base/harness/config/pipeline.yml
+++ b/src/_base/harness/config/pipeline.yml
@@ -7,15 +7,15 @@ command('app build'): |
 
 command('app build console'): |
   #!bash(workspace:/)|@
-  passthru docker build -t @('docker.repository'):@('app.version')-console -f .my127ws/docker/image/console/Dockerfile .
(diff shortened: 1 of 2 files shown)


src/_base/harness/config/pipeline.yml
```diff
@@ -7,15 +7,15 @@ command('app build'): |
 
 command('app build console'): |
   #!bash(workspace:/)|@
-  passthru docker build -t @('docker.repository'):@('app.version')-console -f .my127ws/docker/image/console/Dockerfile .
+  passthru docker build --pull -t @('docker.repository'):@('app.version')-console -f .my127ws/docker/image/console/Dockerfile .
 
 command('app build php-fpm'): |
   #!bash(harness:/docker/image/php-fpm)|@
-  passthru docker build -t @('docker.repository'):@('app.version')-php-fpm .
+  passthru docker build --pull -t @('docker.repository'):@('app.version')-php-fpm .
```
coledave: going to remove the pull here, in static builds there is a reference to the console image which only exists locally and is not part of a registry. as the console image has been pulled the images needed here should already be up to date.

kierenevans: Okay :)

PR #148, opened 2019-08-30
Move docker service environment variable definitions to attributes

andytson-inviqa:
This while supporting still the downstream ability to override _twig/docker-compose.yml/environment.yml

Docker-compose no longer using yaml anchors or it would repeat most env vars twice in the php-fpm service

This is needed so there is no need for an overlay environment.yml in workspaces with duplication of existing env vars when adding to or changing them

For Spryker there is some dynamic env vars, which there is no current attribute way of dynamically defining, so these are staying in the envrronment.yml till something more clever can be done


src/_base/harness/config/functions.yml
```diff
@@ -2,6 +2,10 @@ function('to_yaml', [text]): |
   #!php
   = preg_replace('/^/m', '  ', \Symfony\Component\Yaml\Yaml::dump($text, 100, 2));
 
+function('to_nice_yaml', [text, indentation, nesting]): |
+  #!php
+  = preg_replace('/^/m', str_repeat(' ', $nesting ?: 2), \Symfony\Component\Yaml\Yaml::dump($text, 100, $indentation ?: 2));
+
```
andytson-inviqa, line 8: added as another function as there is no current method to force a ws tool version upgrade, and people will be using versions where optional parameters aren't supported

coledave (approved):

PR #156, opened 2019-10-09
Copy only the web/public directory into Nginx when built in static mode

andytson-inviqa:
Nginx doesn't need the full application.


kierenevans (dismissed):

kierenevans:
I think even magento2 can just have it's /app/pub folder copied in

hgajjar (dismissed):
LGTM

g-foster2020 (approved):

PR #158, opened 2019-10-15
Add basic PVC to persist file uploads for Drupal

g-foster2020:


sgrichards (approved):
LGTM

andytson-inviqa:
A more generic solution I'd already started a few weeks ago https://github.com/inviqa/harness-base-php/pull/161

andytson-inviqa (changes requested):
I'd like consideration for my approach rather than merging this only for mine to be merged as well

g-foster2020:
@andytson-inviqa  That's great in principle but I understand these are needed ASAP yesterday and there is no-one on project to make any changes as we are under pressure to get all the apps deploying to a client environment.

src/drupal8/helm/app/templates/application/php-fpm/deployment.yaml
```diff
@@ -0,0 +1,55 @@
...
+          requests:
+            memory: {{ .Values.resources.requests.memory.php_fpm }}
+        readinessProbe:
+          tcpSocket:
+            port: 9000
+          initialDelaySeconds: 5
+          periodSeconds: 10
+        volumeMounts:
+        - name: php-fpm-persistent-storage
+          mountPath: /app/docroot/sites/default/files
```
andytson-inviqa, line 40: I assume needs to be mounted on the nginx container as well for web access?

I'm not sure about that really, maybe drupal always loads uploaded assets through php endpoints when needed to be web-accessible

andytson-inviqa: I've asked around, and generally nginx does need access. My PR can be used with for example the nfs-server-provisioner chart to provide a test ReadWriteMany storage class

g-foster2020: Thinking it through - we still have the client-level overrides so these changes are not time-dependent; suggest they are changed to general tickets with your comments on how to do it in the better way you have come up with.

PR #159, opened 2019-10-15
Add optional prefix to k8s resources

coledave:


src/_base/helm/qa/values.yaml.twig
```diff
@@ -1,3 +1,4 @@
-app:
+{{ @('workspace.name') }}:
   environment:
     APP_HOST: {{ @('pipeline.qa.hostname') }}
+  prefix: ~
```
andytson-inviqa, line 4: this doesn't override the base app chart's prefix, only prefix: "" does

src/_base/helm/app/values.yaml.twig
```diff
@@ -23,3 +23,4 @@ resources:
 {% include blocks ~ 'resources.yml.twig' %}
 
 ingress: "standard" # standard or istio
+prefix: {{ @('workspace.name') ~ '-' }}
```
andytson-inviqa, line 26: I'd suggest moving this to an attribute and only overriding in the project in question

PR #160, opened 2019-10-16
Add PVCs to Akeneo to persist necessary files

g-foster2020:


andytson-inviqa:
A more generic solution I'd already started a few weeks ago #161

andytson-inviqa (changes requested):
I'd like consideration for my approach rather than merging this only for mine to be merged as well

g-foster2020:
@andytson-inviqa  That's great in principle but I understand these are needed ASAP yesterday and there is no-one on project to make any changes as we are under pressure to get all the apps deploying to a . client environment.

andytson-inviqa:
there's a way to do it in the projects themselves tho using the overlay

PR #161, opened 2019-10-16
Add ability to set up persistent volumes for the application

andytson-inviqa:
Add ability to set up persistent volumes for the application

These will need a ReadWriteMany storageclass available, which k8s clusters generally don't have OTB

If persistence not enabled, then a temporary volume is created for the lifetime of the container
If enabled and no other configuration, then it'll attempt to create a claim the default storageclass, which generally will fail as no support for ReadWriteMany
If enabled and a storageClass is defined, it'll attempt to create a claim on that
If enabled and a hostPath is defined, a HostPath PersistentVolume will be created and a claim made on that

Also supported are selectors for when a persistent volume is already created and can be associated by labels


g-foster2020:
Note akeneo has two file storage locations:
/app/app/file_storage
/app/app/uploads
Maybe that could be reconfigured to something like:
/app/app/volume/file_storage
/app/app/volume/uploads

andytson-inviqa:
or the volume mounted elsewhere and symlinked subfolders.

If it's additions like that, I'd rather it be another PR. Akeneo had no writeable folders in the harness before this

I've added file_storage as a separate volume tho as I saw you have done in a dependent project

PR #162, opened 2019-10-16
Fix db console command name

tkotosz:


kierenevans (dismissed):
Maybe an aliased command for people with muscle memory/shell history?

tkotosz:
@kierenevans updated the PR, I hope this is what you mean. :) Basically the old command will just call the new one.

kierenevans (approved):

PR #163, opened 2019-10-16
Add logs command

tkotosz:


src/_base/harness/config/commands.yml
```diff
@@ -38,6 +38,11 @@ command('exec %'):
   exec: |
     #!bash(workspace:/)|=
     docker-compose exec -T -u build console ={ input.argument('%') }
+    
+command('logs %'):
+  exec: |
+    #!bash(harness:/)|=
+    docker-compose -p ={@('namespace')} logs ={input.argument('%')}
```
kierenevans: Sent `COMPOSE_PROJECT_NAME` env variable in instead of `-p` flag.
I think the absence of this command was to help familiarity with the underlying docker-compose tool, so I'd like @dcole-inviqa's input too :)

tkotosz: updated

andytson-inviqa: @kierenevans yep tho COMPOSE_PROJECT_NAME uses the workspace.name, whereas docker-compose normally uses the folder name, so can differ. Personally I'd prefer the folder name to be used, but is outside of this PR

g-foster2020 (approved):

PR #175, opened 2019-11-21
Make entrypoint.sh a true entrypoint and use sleep infinity

andytson-inviqa:


coledave (dismissed):

src/_base/docker/image/console/root/entrypoint.sh
```diff
@@ -24,4 +18,7 @@ bootstrap()
 }
 
 bootstrap
-main
+
+if [ "$#" -gt 0 ]; then
+    exec "$@"
+fi
```
andytson-inviqa, line 26: bug here in that "sleep" isn't a cli app but a bash function. Easy to fix tho

PR #185, opened 2019-11-26
Merge the nginx and php-fpm deployments together as a webapp pod

andytson-inviqa:
* Also remove php-fpm k8s service as it's never used external to the pod now
* Update FPM_HOST to localhost for local pod communication


src/spryker/helm/app/templates/application/webapp/ingress.yaml
```diff
@@ -5,16 +5,16 @@ kind: Ingress
...
   creationTimestamp: null
   labels:
-    app.service: {{ .Values.resourcePrefix }}nginx
-  name: {{ .Values.resourcePrefix }}nginx
+    app.service: {{ .Values.resourcePrefix }}webapp
+  name: {{ .Values.resourcePrefix }}webapp
 spec:
   rules:
   {{- range $key, $value := .Values.docker.services.nginx.environment }}
   - host: {{ $value | quote }}
```
andytson-inviqa, line 13: FPM_HOST: localhost in spryker here will be weird (just as it was wierd FPM_HOST: php-fpm before). it shouldn't really use the env vars here anyway tho, which is a hack to reduce duplication

g-foster2020 (dismissed):

src/_base/helm/app/templates/application/webapp/deployment.yaml
```diff
@@ -1,22 +1,22 @@
...
   template:
     metadata:
       labels:
-        app.service: {{ .Values.resourcePrefix }}php-fpm
+        app.service: {{ .Values.resourcePrefix }}webapp
     spec:
       {{- if not (eq "" (include "application.volumes.wwwDataPaths" .)) }}
       initContainers:
-      - name: php-fpm-volume-permissions
+      - name: {{ .Values.resourcePrefix }}webapp-volume-permissions
```
coledave: for container names we can get away without the resourcePrefix, in the case of multiple containers it makes knowing which container to specify a bit more consistent

src/_base/helm/app/templates/application/webapp/deployment.yaml
```diff
@@ -31,7 +31,39 @@ spec:
         {{- end }}
       {{- end }}
       containers:
-      - env:
+      - name: {{ .Values.resourcePrefix }}nginx
```
coledave: no need for the resource prefix

src/_base/helm/app/templates/application/webapp/deployment.yaml
```diff
@@ -31,7 +31,39 @@ spec:
...
+            port: 80
+          initialDelaySeconds: 5
+          periodSeconds: 10
+        volumeMounts:
+        {{- if not (eq "" (include "application.volumeMounts.all" .)) }}
+        {{- include "application.volumeMounts.all" . | indent 8 }}
+        {{- else }} []
+        {{- end }}
+
+      - name: {{ .Values.resourcePrefix }}php-fpm
```
coledave: no need for resource prefix

andytson-inviqa: ah ok, this originally had the prefix on (just moved up to the top of the object to make it more visible in a multi-container pod definition), so I added it in the other containers of the pod too

Removed here now

src/_base/helm/app/templates/application/webapp/deployment.yaml
andytson-inviqa: removed here now

src/_base/helm/app/templates/application/webapp/deployment.yaml
andytson-inviqa: removed here now

coledave (approved):

PR #188, opened 2019-11-28
[DEVOPS-1115] just the build changes

g-foster2020:


src/spryker/docker/image/console/root/lib/task/spryker/install.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+function task_spryker_install()
+{
+    passthru "PGPASSWORD=$DB_PASS vendor/bin/install -r docker -x frontend"
+    passthru "PGPASSWORD=$DB_PASS vendor/bin/install -r docker -s database-migrate"
+    passthru "PGPASSWORD=$DB_PASS vendor/bin/install -r docker -s demodata"
```
andytson-inviqa: do we always need the demo data? Any use beyond getting a base sample environment? Does it need to be removed when real data is provided?

g-foster2020: I'll make it optional, off by default.

g-foster2020:
Happier with that now - I have tried to keep both the steps we found necessary in other project, and H's harness steps, without any duplication.

It may be that, later when in actual use, we find certain steps we may want to be optional.  We can make further improvements then.

andytson-inviqa (approved):
Looks ok to me, but I don't know for certain which spryker tasks should be run when, just when they shouldn't be

PR #189, opened 2019-12-04
Avoid rendering postgres password in debug output

andytson-inviqa:


andytson-inviqa:
nvm, mysql is doing it as well, and need a better way to solve that

PR #193, opened 2019-12-09
Pass MySQL server variables

bennoislost:
Initial attempt at configuring MySQL passing container command arguments for transparency.


bennoislost:
Will produce further PR for Helm configuration

PR #195, opened 2019-12-11
[DEVOPS-1119] fix for spryker build and deployment

g-foster2020:


andytson-inviqa (commented):
I don't know the reason for assets:dump being needed

g-foster2020:
> I don't know the reason for assets:dump being needed

I believe it is so that, on a local rebuild rather than having to create the database again by scripts being executed, it is simply loaded directly from the database dump.  As this is the place to put it, and it was being used in this way in spryker previous to this update, I thought it best to demonstrate where it would go but comment it out as it's not a required step, as per demodata step.

andytson-inviqa:
ok, might be best that is only added when in local environments then

andytson-inviqa:
sorry accidentally closed

PR #200, opened 2019-12-20
Add database import command and steps configured in harness.yml

g-foster2020:


src/akeneo/harness.yml
```diff
@@ -30,6 +30,17 @@ attributes:
     user: akeneo
     pass: akeneo
     name: akeneo
+    import:
+      steps:
+        - passthru "echo \"Importing database file\""
+        - run "set -e -o pipefail; zcat \"$1\" | mysql -h \"$DB_HOST\" -u root -p\"$DB_ROOT_PASS\" \"$DB_NAME\""
```
kierenevans: Missing the `pv` pipe via passthru - important for large databases.
https://github.com/inviqa/harness-base-php/blob/0.3.x/src/_base/docker/image/console/root/lib/task/assets/apply.sh#L28 has an example

src/akeneo/harness.yml
```diff
@@ -30,6 +30,17 @@ attributes:
     user: akeneo
     pass: akeneo
     name: akeneo
+    import:
+      steps:
+        - passthru "echo \"Importing database file\""
```
kierenevans: I would use single quotes if there's no variables inside the string

src/akeneo/harness.yml
kierenevans: Assumption of root user will conflict with https://github.com/inviqa/harness-base-php/pull/197

kierenevans (changes requested):

src/_base/harness/config/commands.yml
```diff
@@ -202,3 +202,9 @@ command('feature xdebug cli (on|off)'):
     run ws service php-fpm restart
     echo 'Done'
 
+command('db import <database_file>'):
+  env:
+    COMPOSE_PROJECT_NAME: = @('namespace')
+  exec: |
+    #!bash(workspace:/)|@
```
kierenevans: To use `={input.argument('database_file')}`, this needs to be:
```suggestion
    #!bash(workspace:/)|=
```

src/akeneo/harness.yml
g-foster2020: Yes some confusion there, I think there is more work to be done for #197 as there are still scripts in the harnesses using DB_ROOT_PASS, either that or I somehow don't have these changes.... I will check and remove DB_ROOT_PASS and convert anything else as per #197

g-foster2020: Seems just my confusion.... Andy left in DB_ROOT_PASS I guess for backwards compatibility and misread this

src/_base/harness/config/commands.yml
g-foster2020: The thing is.... it does work?

g-foster2020: So something wrong with the ws code expected result.  I will still change it as I can see the intention.

kierenevans (approved):

PR #201, opened 2020-01-10
Allow additional istio gateways to be specified

g-foster2020:


src/_base/helm/app/templates/application/webapp/ingress-istio.yaml
```diff
@@ -11,6 +11,11 @@ spec:
  - {{ index .Values.docker.services "php-base" "environment" "APP_HOST" }}
  gateways:
  - istio-system/{{ .Release.Namespace }}-gateway
+{{- if index . "Values.istio.additionalGateways" }}
```
andytson-inviqa: I don't think index traverses dots

andytson-inviqa: well it does, but I've only seen it do that when doing `index .Values "istio" "additionalGateways", but that wont check if .Values.istio exists so would fail too

g-foster2020: hmm

g-foster2020: how about using with combined with if index

"with skips the code block if the variable is absent from the template"

g-foster2020: or alternatively, ensure .Values.istio[.additionalGateways] = [] is defined in the base config

andytson-inviqa: yes latter is cleaner

andytson-inviqa (approved):

PR #203, opened 2020-01-14
MISC Pass script arguments into function for further use

g-foster2020:


src/spryker/docker/image/jenkins-runner/root/entrypoint.jenkins_runner.dynamic.sh
```diff
@@ -12,4 +12,4 @@ bootstrap()
 }
 
 bootstrap
-main
+main $@
```
kierenevans: Is shellcheck happy with this?

g-foster2020: No, its not (never heard of it until you mentioned it) - I will update.

andytson-inviqa (approved):

PR #204, opened 2020-01-16
Support use of specific mode in akeneo commands

g-foster2020:


src/akeneo/docker/image/php-fpm/root/etc/supervisor/conf.d/akeneo-job-queue-consumer.conf
```diff
@@ -0,0 +1,9 @@
+[program:akeneo-job-queue-consumer]
+command=php /app/bin/console --env=%(ENV_AKENEO_APP_MODE) akeneo:batch:job-queue-consumer-daemon
+stdout_logfile=/dev/stdout
+stdout_logfile_maxbytes=0
+stderr_logfile=/dev/stderr
+stderr_logfile_maxbytes=0
+user=www-data
+autostart=%(ENV_AUTOSTART_AKENEO-JOB-QUEUE-CONSUMER)s
```
kierenevans: dashes?

g-foster2020: Not my choice - it's always been called that, this is an override file that should already be in the harness as its required for the akeneo to function.  So without this the harness is functionally incomplete.

kierenevans:
```
export ENV_AUTOSTART_AKENEO-JOB-QUEUE-CONSUMER=1
bash: export: `ENV_AUTOSTART_AKENEO-JOB-QUEUE-CONSUMER=1': not a valid identifier
```

Needs fixing, not valid in bash.

kierenevans (approved):

PR #205, opened 2020-01-21
bug fix for intended config of additional gateways

g-foster2020:


andytson-inviqa (approved):

andytson-inviqa (changes requested):
Doesn't match the values.yaml

PR #206, opened 2020-01-27
Akeneo - split job-queue-consumer from php fpm container

g-foster2020:
Currently the php-fpm also runs the job-queue-consumer.  It is also a good idea for the number of job-queue-consumers to be configurable.


src/akeneo/_twig/docker-compose.yml/application.yml.twig
```diff
@@ -0,0 +1,71 @@
...
+{% else %}
+    image: {{ @('docker.repository') ~ ':' ~ @('app.version') ~ '-php-fpm' }}
+{% endif %}
+    labels:
+      - traefik.enable=false
+    networks:
+      - private
+    environment:
+{{ deep_merge_to_yaml([@('services.php-base.environment'), @('services.php-fpm.environment')], 2, 6) | raw }}
+{% include blocks ~ 'environment.yml.twig' %}
```
andytson-inviqa, line 71: feels a bit "spot the difference between this and _base's one". I can't see anything of significance

src/akeneo/harness/config/q-akeneo-pipeline.yml
```diff
@@ -0,0 +1,21 @@
+
```
andytson-inviqa, line 1: why q-akeneo-pipeline.yml?

andytson-inviqa: a before p before q I guess

src/akeneo/_twig/docker-compose.yml/application.yml.twig
g-foster2020: it's the image name: 'image: {{ @('workspace.name') ~ '-php-fpm:dev' }}' so it can be used as the basis for the job-queue-consumer Dockerfile which only differs by one file (the supervisor config)

src/akeneo/harness/config/q-akeneo-pipeline.yml
g-foster2020: yes... spryker-pipeline.yml was nicer... but the right place in the alphabet

src/akeneo/_twig/docker-compose.yml/application.yml.twig
g-foster2020: I copied this approach from spryker and jenkins-runner, IIRC

g-foster2020: e.g. see console:dev https://github.com/inviqa/harness-spryker/blob/0.3.x/_twig/docker-compose.yml/application.yml.twig

andytson-inviqa (approved):

src/akeneo/_twig/docker-compose.yml/application.yml.twig
andytson-inviqa: ok. I'd be curious if we can do something more standard later that could fit in the base. I have a few ideas on simplification of build process that might help

PR #210, opened 2020-02-05
Add cron container, with it enabled for Akeneo 

g-foster2020:


src/akeneo/harness/config/confd.yml
```diff
@@ -1,6 +1,5 @@
 
 confd('harness:/'):
-  # base templates
```
andytson-inviqa: can we keep this?

andytson-inviqa (dismissed):

src/_base/docker/image/cron/root/cron-run-with-env.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+IFS=$'\n'
+env_vars="$(cat /app/env.sh)"
+/usr/bin/env - "$env_vars" "$@" > /proc/1/fd/1 2> /proc/1/fd/2
```
andytson-inviqa: does this definitely work correctly with multiple env vars? with the now added quotes, I believe it would see as
```
/usr/bin/env - "ENV_VAR1=value1 ENV_VAR2=value2" ...
```
rather than:
```
/usr/bin/env - "ENV_VAR1=value1" "ENV_VAR2=value2" ...
```

This is why I mentioned you may need to tell shellcheck to ignore the lack of quoting

g-foster2020: yes it works

g-foster2020: no.. maybe not! I will adjust it

g-foster2020: I will re-test it to be sure

g-foster2020: The following cron 

* * * * * /cron-run-with-env.sh env

Gives me this in the logs

❯ docker logs b500f2b8532a                                                                                                12s
PROCESS_ENDPOINT=xxxxxxxx
...
PHP_MD5=
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/app/bin
...
_=/usr/bin/env

BUT ... running the crontab under the www-data user seems to prevent the logs from being appended to the root process's stdout, so I will adjust how that works

g-foster2020: I can't run crontab directly under www-data as it cannot log to stdout of the main process for docker logs to view.  So I have changed the script used in the crontab to run commands as www-data.  Not as good, but we need the logs.

andytson-inviqa: yep, makes sense actually, as you're piping the output to the /proc/1, which no non-root user would be able to do. The change to run just the internal task as the user will be good enough still, as that's where the main concern arrises

src/spryker/harness/scripts/enable.sh.twig
```diff
@@ -34,8 +34,14 @@ dynamic()
...
-    passthru "docker-compose config --services | grep -v console | xargs docker-compose pull"
-    passthru "docker-compose config --services | grep -v jenkins-runner | xargs docker-compose build --pull"
+    {% if ("cron" in @('app.services')) %}
+        passthru "docker-compose config --services | grep -v php-fpm | grep -v console | xargs docker-compose pull"
+        passthru "docker-compose config --services | grep -v cron | grep -v jenkins-runner | xargs docker-compose build --pull"
+        passthru docker-compose build cron
+    {% else %}
+        passthru "docker-compose config --services | grep -v console | xargs docker-compose pull"
+        passthru "docker-compose config --services | grep -v jenkins-runner | xargs docker-compose build --pull"
+    {% endif %}
```
andytson-inviqa, line 44: I'd like to maybe later move the if condition to service env var so it doesn't need twig rendering, but ok for now.

src/akeneo/harness/scripts/enable.sh.twig
```diff
@@ -52,6 +53,7 @@ static()
     ws app build
 
     passthru "docker-compose config --services | grep -v job-queue-consumer | xargs docker-compose up -d"
+    passthru sleep 60s # ensure mysql has had time to start
```
andytson-inviqa: can we change this to do tcp retry checks instead? having to always wait 60 seconds for mysql to be started is very inefficient. I can accept it as a separate PR tho when someone has some time

andytson-inviqa (dismissed):

src/akeneo/harness/scripts/enable.sh.twig
g-foster2020: this is no longer need so I will remove it

g-foster2020:
@andytson-inviqa I removed the sleep line if you could re-approve, thanks

andytson-inviqa (approved):

PR #211, opened 2020-02-10
Fix cron script to work with php commands + Spryker timeout issue + Build fixes/improvements

g-foster2020:
For whatever reason, 

`/usr/bin/env - "$env_vars" sudo --preserve-env -u www-data "$@" `

works with env, but not with php commands (e.g. /bin/console) - for some reason they "preserved environment variables" do not end up in the $_ENV array.


andytson-inviqa (dismissed):

kierenevans (approved):

kierenevans:
I've approved in the interests of time, but perhaps instead of shipping whole files from Spryker as overrides, could we ship a composer patch?

g-foster2020:
> I've approved in the interests of time, but perhaps instead of shipping whole files from Spryker as overrides, could we ship a composer patch?

They aren't from Spryker, Hardik created them... but the Spryker code sniffer requires all the files to have the file doc block at the top, which is misleading but hopefully these files are only temporary (bug has been raised with Spryker).

PR #215, opened 2020-02-21
Add mutagen sync

hgajjar:


hgajjar:
How should we test the mutagen sync? By default I have kept it disabled, so `./test <harness> dynamic` will only test it with docker-sync. 
Should we extend this `test` script to allow testing with both docker-sync and mutagen options? Ex. `./test harness dynamic mutagen`, which will then enable mutagen into workspace.override.yml and test the environment?

kierenevans:
https://github.com/inviqa/harness-base-php/pull/230 targets this PR as merge target, adds testing and gets docker-sync in line with the mutagen approach :)

kierenevans (approved):
Amazing work!

PR #218, opened 2020-02-24
Memory requests equal limits

kierenevans:
Built on top of #217 to catch the cron memory moving.


andytson-inviqa (dismissed):
while I think its ok to have the requests match limits, I don't think it should be forced like this

README.md
```diff
@@ -18,6 +18,41 @@ Each framework will fully override a base harness file if differing behaviour is
...
+* Skeleton for simple set-up of new projects
+* Pipeline docker-compose environment for use in Jenkins or other tools to run tests
+* Helm chart for deploying QA, UAT and Production environments to Kubernetes clusters
+
+## Helm charts
+
+Each harness deploys:
+* A "console" pod for running one-off commands
+* A NGINX/PHP-FPM "webapp" pod for running the php-based application and serving web requests
+* A "cron" pod for running cronjobs
```
andytson-inviqa: optional

andytson-inviqa (commented):
If we really want to force it, then it'll be less confusing to drop the `limit` from the keys too

README.md
kierenevans: Updated

kierenevans:
This is number one source of unstable nodes that I've seen over the years, so I think it's super important we don't let people make this mistake when it's avoidable.

kierenevans:
Rebased, fingers crossed for the build.

coledave (dismissed):

coledave (approved):

PR #223, opened 2020-02-26
Add php-fpm-exporter to provide a prometheus endpoint

g-foster2020:


src/_base/helm/app/templates/service/php-fpm-exporter/deployment.yaml
```diff
@@ -0,0 +1,32 @@
...
+  selector:
+    matchLabels:
+      app.service: {{ .Values.resourcePrefix }}php-fpm-exporter
+  template:
+    metadata:
+      labels:
+        app.service: {{ .Values.resourcePrefix }}php-fpm-exporter
+    spec:
+      containers:
+      - image: hipages/php-fpm_exporter
```
kierenevans: Should this not be in the same pod as nginx/php-fpm ("web" deployment)
Otherwise php-fpm would need to be exposed to the network.

g-foster2020: Good point.

src/_base/helm/app/templates/application/webapp/service.yaml
```diff
@@ -13,3 +13,21 @@ spec:
     app.service: {{ .Values.resourcePrefix }}webapp
 status:
   loadBalancer: {}
+{{ if .Values.service.php_fpm_exporter }}
```
kierenevans: Does helm like this two objects in one for v3? I had errors in v2 when doing this elsewhere

src/_base/helm/app/templates/application/webapp/deployment.yaml
```diff
@@ -106,6 +106,24 @@ spec:
...
+      
+      {{ if .Values.service.php_fpm_exporter }}
+      - name: php-fpm-exporter
+        image: hipages/php-fpm_exporter
+        imagePullPolicy: Always
+        ports:
+        - containerPort: 9253
+        resources:
+          limits:
+            memory: {{ .Values.resources.limits.memory.php_fpm_exporter }}
```
kierenevans, line 118: Missing some defaults for this and the requests below.

src/_base/helm/app/templates/application/webapp/service.yaml
g-foster2020: I'm sure we already do it somewhere. Hmm. Trying to think where.

g-foster2020: No maybe not, I will split them just to be safe.

kierenevans:
Don't forget spryker `resources.yml.twig` and any other overrides per framework

g-foster2020:
> Don't forget spryker `resources.yml.twig` and any other overrides per framework

Good point, I'll review them all. Not used to adding things to _base.

g-foster2020:
These are the files that are modified in _base that may have overrides:

- □ src/_base/harness/attributes/common.yml
- □ src/_base/helm/app/_twig/values.yaml/resources.yml.twig
- □ src/_base/helm/app/values.yaml.twig

g-foster2020:
@kierenevans I think that's everything now?

kierenevans (approved):

PR #229, opened 2020-03-06
Fix akeneo production mode

g-foster2020:


kierenevans:
I think this also needs dev dependencies to get installed when not in "development" mode: https://github.com/inviqa/harness-base-php/blob/0.4.x/src/akeneo/docker/image/console/root/lib/task/composer/install.sh.twig#L8

g-foster2020:
> I think this also needs dev dependencies to get installed when not in "development" mode: https://github.com/inviqa/harness-base-php/blob/0.4.x/src/akeneo/docker/image/console/root/lib/task/composer/install.sh.twig#L8

What we do is, in Jenkinsfile build we run the composer dev install before tests run.  That way we avoid having the dev dependencies in the production images.

PR #230, opened 2020-03-07
Mutagen and Docker Sync Testing

kierenevans:
Adds testing for mutagen during the dynamic build.

If we install GCC or other ruby-build dependencies to the Jenkins runners, we can get docker-sync testing too.

docker-sync usage refactored to be like the new mutagen implementation.
docker-sync also installing and using rbenv instead of relying on the system ruby (caused issues for some projects where devs had not used hem before).

### DIFF
--- a/src/_base/docker/image/console/root/entrypoint.dynamic.sh
+++ b/src/_base/docker/image/console/root/entrypoint.dynamic.sh
@@ -31,6 +31,8 @@ resolve_volume_mount_strategy()
             STRATEGY="host-osx-normal"
         elif (mount | grep "/app type ext4") > /dev/null 2>&1; then
             STRATEGY="host-osx-dockersync"
+        elif (mount | grep "/app type btrfs") > /dev/null 2>&1; then
+            STRATEGY="host-linux-normal"
         else
             exit 1
         fi
--- a/src/_base/docker/image/php-fpm/root/entrypoint.dynamic.sh
+++ b/src/_base/docker/image/php-fpm/root/entrypoint.dynamic.sh
@@ -46,6 +46,8 @@ resolve_volume_mount_strategy()
             STRATEGY="host-osx-normal"
         elif (mount | grep "/app type ext4") > /dev/null 2>&1; then
             STRATEGY="host-osx-dockersync"
+        elif (mount | grep "/app type btrfs") > /dev/null 2>&1; then
+            STRATEGY="host-linux-normal"
         else
             exit 1
         fi


src/_base/mutagen.yml.twig
```diff
@@ -1,10 +1,10 @@
 forward:
-  teufel:
+  {{ @('workspace.name') }}:
```
hgajjar: 🤦‍♂ , thanks for fixing this, I totally missed it

src/_base/harness/scripts/mutagen.sh
```diff
@@ -1,31 +1,54 @@
...
+    if command -v mutagen > /dev/null 2>&1; then
+        return 0
     fi
+
+    if command -v sw_vers > /dev/null 2>&1 && sw_vers | grep -q Mac; then
+        if command -v brew > /dev/null 2>&1; then
+            passthru brew install mutagen-io/mutagen/mutagen
+        else
+            mkdir -p .my127ws/utilities/mutagen/
+            run curl -L -q -sS -f https://github.com/mutagen-io/mutagen/releases/download/v0.11.2/mutagen_darwin_amd64_v0.11.2.tar.gz -o .my127ws/utilities/mutagen/mutagen.tar.gz
```
hgajjar: Is there a way to get latest release tag, so we always download latest version? 🤔

test
```diff
@@ -13,24 +15,66 @@ function test()
...
     local mode="$2"
 
+    setup "$harness" "$mode"
+    setup_dynamic_mountpoint
+    install_environment
+    check_environment_started "$harness" "$mode"
+    run_pipeline_tests
+    restart_environment
+    check_environment_started "$harness" "$mode"
+    teardown
```
hgajjar: Should this be inside if mode == static?

hgajjar (commented):
Overall it looks good to me, just few comments

test
kierenevans: It handles the default version of dynamic too (mountpoints on linux)

hgajjar: aha, okay

kierenevans: I'll add a comment to explain :)

src/_base/harness/scripts/latest-mutagen-release.php
```diff
@@ -0,0 +1,33 @@
...
+
+$osFamily = strtolower(PHP_OS_FAMILY);
+$releaseAssets = array_filter($latestStableRelease['assets'], function ($asset) use ($osFamily) {
+  return preg_match("/^mutagen_.*${osFamily}.*amd64.*/", $asset['name']) > 0;
+});
+$releaseAsset = reset($releaseAssets);
+if (!$releaseAsset) {
+  throw new \Exception('Could not find latest stable release asset for your OS Platform: ' . $osFamily);
+}
+echo $releaseAsset['browser_download_url'];
```
hgajjar: Awesome 🎉

hgajjar (approved):
Looks really good, however pr build is failing

kierenevans:
Yeah testing locally on linux machine and the mountpoint isn't mounting.
UPDATE: That's separate as it worked on the Jenkins runner. Was a PHPCS Failure for spryker harness (which is looking at all PHP files including .my127ws)

kierenevans:
Testing now working fine - it was due to my docker configuration using brtfs mountpoints :)

src/_base/mutagen.yml.twig
kierenevans: NP :)

PR #241, opened 2020-03-17
Add volume mounts to the init and migrate pod specs

coledave:


kierenevans (approved):

src/_base/helm/app/templates/application/app-init.yaml
```diff
@@ -45,7 +45,18 @@ spec:
             memory: {{ .Values.resources.limits.memory.app_init }}
           requests:
             memory: {{ .Values.resources.requests.memory.app_init }}
+        volumeMounts:
+        {{- if not (eq "" (include "application.volumeMounts.all" .)) }}
+        {{- include "application.volumeMounts.all" . | indent 8 }}
```
andytson-inviqa, line 50: need to also include backend here

src/_base/helm/app/templates/application/app-migrate.yaml
```diff
@@ -46,7 +46,18 @@ spec:
             memory: {{ .Values.resources.limits.memory.app_migrate }}
           requests:
             memory: {{ .Values.resources.requests.memory.app_migrate }}
+        volumeMounts:
+        {{- if not (eq "" (include "application.volumeMounts.all" .)) }}
+        {{- include "application.volumeMounts.all" . | indent 8 }}
```
andytson-inviqa, line 51: need to also include backend here

andytson-inviqa:
note this PR makes it necessary to use ReadWriteMany storage by default, and means it's not able to choose not to (console had an option "persistence.mountVolumesOnConsole" to be turned off)

PR #249, opened 2020-03-23
pre-pull docker images

coledave:
As deployments roll out the synchronous way images are pulled at each step can make it quite slow. This is just an idea to pull images in advance across nodes.


kierenevans:
In a multiple application cluster, all nodes won't be involved in a single application's hosting. This could case space issues on some nodes where the image isn't needed, triggering disk cleanup sooner (node might run out of inodes or space).

PR #263, opened 2020-03-28
Update Magento 2 application skeleton - DEV tools configuration

tkotosz:
Changelog:
1. Use [BehatMagento2Extension](https://github.com/tkotosz/BehatMagento2Extension) instead of te deprecated Behat extensions. Removed extensions:
- [Magento2InitExtension](https://github.com/tkotosz/behat-magento2-init): Not needed when we use BehatMagento2Extension
- [FOB ContextServiceExtension](https://github.com/FriendsOfBehat/ContextServiceExtension): Become deprecated a long time ago and BehatMagento2Extension can provide the same functionality
- [FOB CrossContainerExtension](https://github.com/FriendsOfBehat/CrossContainerExtension): Become deprecated a long time ago and BehatMagento2Extension can provide the similar functionality
2. Remove "friendsofphp/php-cs-fixer": phpcbf is included in the phpcs package and usually we use that for fixing phpcs reported issues
3. Add "fix-code" command to composer scripts to alias phpcbf
4. Update php linting script to only output errors
5. Add missing phpspec configuration: Set the formatting to the usual pretty one and load the phpspec Magento 2 extension
6. Add missing, regularly used InviqaNonComposerComponentRegistration.php to be able to load Magento modules from src + update gitignore to allow pushing this file + caching implemented for it
7. Replace deprecated psr-0 autoload with psr-4
8. Normalize composer json formatting with [composer normalizer](https://packagist.org/packages/ergebnis/composer-normalize)
9. Add missing php minimum version requirement to composer config
10. Add Behat example to verify Behat and Magento can communicate properly (dependency injection from Magento to Behat)
11. Update phpstan config to ignore registration.php files


src/magento2/application/skeleton/app/etc/InviqaNonComposerComponentRegistration.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+$pathList[] = realpath(__DIR__ . '/../..') . '/src/*/*/cli_commands.php';
+$pathList[] = realpath(__DIR__ . '/../..') . '/src/*/*/registration.php';
+
+foreach ($pathList as $path) {
+    // Sorting is disabled intentionally for performance improvement
+    $files = glob($path, GLOB_NOSORT);
```
kierenevans: This is slow and runs filesystem scanning on each request - can we optimise to have a module manifest?

tkotosz: ah ye I remember you did some optimisation for this on toom

src/magento2/application/skeleton/behat.yml.twig
```diff
@@ -1,40 +1,10 @@
...
-    SensioLabs\Behat\PageObjectExtension:
-      namespaces:
-        page: [Acceptance\Page]
-    FriendsOfBehat\ContextServiceExtension:
-      imports:
-        - 'features/bootstrap/_config/contexts.yml'
-    FriendsOfBehat\CrossContainerExtension: ~
-    Cjm\Behat\StepThroughExtension: ~
-    DMore\ChromeExtension\Behat\ServiceContainer\ChromeExtension: ~
-local: # (google-chrome-stable --remote-debugging-port=9221 &) && socat tcp-l:9222,fork TCP4:127.0.0.1:9221
```
kierenevans, line 36: We should keep this config

src/magento2/application/skeleton/app/etc/InviqaNonComposerComponentRegistration.php
tkotosz: hmm and I just noticed that we could get rid of this file because it seems now magento has a separate file (registration_globlist.php) which we could put under version control to add the "src" dir to it... probably that would be the default way of doing this nowadays

for the improvement which you did on toom I think that should be something which replaces the default NonComposerComponentRegistration.php provided by magento :think: (maybe they would accept a PR for that in the upstream repo)

tkotosz: ah ye I remember you did some optimisation for this one of the projects

hmm and I just noticed that we could get rid of this file because it seems now magento has a separate file (registration_globlist.php) which we could put under version control to add the "src" dir to it... probably that would be the default way of doing this nowadays

for the improvements which you mentioned I think should be something which replaces the default NonComposerComponentRegistration.php provided by magento :think: (maybe they would accept a PR for that in the upstream repo)

src/magento2/application/skeleton/behat.yml.twig
tkotosz: okay, I had no idea what are those :D (removed it together with all the browser based stuff)

tkotosz: done, Mink is available again, tested and works with goutte and chrome as well

src/magento2/application/skeleton/app/etc/InviqaNonComposerComponentRegistration.php
tkotosz: added the registration_globlist.php now... I tested it and works, but I am not sure if it is overwrited during the inital project setup or not... will do some testing around that. :)

tkotosz: well it got overridden probably during composer install :/ will try it with the harness overlay to see if that fixes it

tkotosz: well that didn't work either :/

tkotosz: ok I will go back to InviqaNonComposerComponentRegistration.php and also add the caching stuff (since registration_globlist overwrited by composer on every magento upgrade so it is not ment to be under version control for sure)

PR #281, opened 2020-04-27
Adding documentation for setting up workspace on Drupal projects

sawtell:
This PR adds documentation for the Drupal 8 harness.

The documentation covers:
- Installing a new project, existing project and a multi-site project
- Customising the Workspace environment with overrides and helper commands
- Advanced customisation for adding services and using twig templates
- Some information on how to contribute to the project. (This could be improved)


src/drupal8/docs/customise/commands.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+
+### Frontend development - `ws frontend-start`
+```yaml
+attribute('frontend.path'): "/app/docroot/themes/custom/mytheme"
+
+# ...
+
+command('frontend-start'):
```
kierenevans: I think this could be adapted to be `ws frontend watch` https://github.com/inviqa/harness-base-php/blob/0.6.x/src/_base/harness/config/commands.yml#L135-L140

Need to set `attribute('frontend.watch'): 'yarn dev'` and it will work with the frontend.path attribute above

src/drupal8/docs/customise/commands.md
```diff
@@ -0,0 +1,42 @@
...
+    ws composer install
+    ws drush -y cr
+    ws drush -y updb
+    ws frontend install
+```
+
+
+### Database import - `ws import-db path/to/db.sql`
+```yaml
+command('drush-import-db <alias> <filename>'):
```
kierenevans: I think this can be made into the `ws db import <filename>` here: https://github.com/inviqa/harness-base-php/blob/0.6.x/src/_base/harness/config/commands.yml#L212-L217 - maybe an overridden version for the drupal8 harness?

kierenevans:
This is really amazing to see. Thank you!

src/drupal8/docs/customise/commands.md
sawtell: Thanks, I've removed this to avoid confusion with the commands that are already defined.

src/drupal8/docs/customise/commands.md
sawtell: I thought I would leave this example in, just because the base import command doesn't account for multiple database set ups. Happy to change though.

src/drupal8/docs/install/multi-site.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Creating multiple databases during install
+_If you are using table prefixes rather than multiple databases then you can skip this section._
+
+In order to use multiple databases in single mysql container you will need to change the defaults to use the continuous pipe mysql image.
+
+In `workspace.yml` add:
+```yaml
+attribute('mysql.image'): quay.io/continuouspipe/mysql5.7
+attribute('mysql.tag'): latest
```
sawtell: @kierenevans, I've left this as `latest` for now. Will the continouspipe image be updated or should I change this to either use `stable` or add the grants in the install steps?

kierenevans: Let's update to use the init step grants instead of the CP image

src/drupal8/docs/customise/commands.md
kierenevans: Tis okay :)

src/drupal8/docs/customise/commands.md
kierenevans: Ace!

src/drupal8/docs/install/multi-site.md
sawtell: Sure - my concern with that approach is that they don't trigger once you have a DB dump in `tools/assets`. Probably not an issue for the initial install, but if someone destroys and re-installs then the databases won't be created.

Taking a look at other options, would it work if we created them in the `before('harness.install')` event?

kierenevans: I think it's triggered too early to use against a running container: https://github.com/my127/workspace/blob/0.1.x/src/Types/Workspace/Installer.php#L79

The `ws enable` command is called by the `after.harness.install` event triggered here:
https://github.com/my127/workspace/blob/0.1.x/src/Types/Workspace/Installer.php#L97
and received here: https://github.com/inviqa/harness-base-php/blob/17d88d3ba9a48b1715211ee81307284d70914ce7/src/_base/harness/config/events.yml#L2-L4

Only then is the set of docker containers brought up.

After that we are left with either build step, install step, init step, or customising `enable.sh` to be able to do extra things in between.

kierenevans: > Sure - my concern with that approach is that they don't trigger once you have a DB dump in `tools/assets`.

I think that's why we did the grants as the init step rather than install step.

src/drupal8/docs/install/multi-site.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+The database dumps should be located in `tools/assets/development`, ensuring that the filename is the same as the database name that you defined in the previous step:
+```
+project-root/
+├- ...
+└- tools/
+.  └- assets/
+.  .  └- development/
+.  .  .  ├- drupal.sql.gz
```
michaeltrestianu18: Needs updated to `one_site.sql.gz`

src/drupal8/docs/install/multi-site.md
sawtell: Updated to use the init steps.

src/drupal8/docs/install/existing-project.md
```diff
@@ -0,0 +1,41 @@
...
+    ```
+    cd /path/to/project
+    ```
+    From here, run the `ws create` command, but don't run the installation:
+    ```
+    ws create <projectName> inviqa/drupal8:v0.9.4
+    ```
+    (Note: see [harnesses.json] for available versions.)
+
+2. You will now have a new directory, `<projectName>` containing a number of files and folders. We don't need to be concerned with all of them.
```
kierenevans, line 13: Yes we do :D

kierenevans:
@sawtell are you okay to review my changes in the last 7 commits? https://github.com/inviqa/harness-base-php/pull/281/files/ca6adca9db84d5a7335279207d657bdf5fe67ced..d06b37f0861daf6debd5ff493c2dd0273cbcfea6

Thanks!

sawtell:
Hi @kierenevans, I can't approve them, but they look great! Thanks for fixing the bugs and improved explanations. 👍

kierenevans (approved):

PR #287, opened 2020-05-06
Akeneo 4.x Support

kierenevans:
* App skeleton no longer required due to akeneo post-install script that copies files into place
* Requirements for php, mysql and elasticsearch versions met
* Initial support for choosing enterprise edition


a-ast (dismissed):
Great work!

a-ast (approved):
Great job ;)

PR #292, opened 2020-05-11
Ensure latest drupal/core 8.8.5 is installed. Fix test-quality script

kierenevans:
`webflo/drupal-core-require-dev` is deprecated and for some reason wasn't pulling in the drupal/coder package to the right place.

Replace with the recommended replacement `drupal/core-dev`.

Upgrade `drupal/pathauto` module so it brings in latest `drupal/core` package.


sawtell (approved):
LGTM 👍

PR #296, opened 2020-05-13
Add missing php-fpm pool overrides to pool config

andytson-inviqa:
This only affected pools other than www, as the www pool has this additional configuration in /usr/local/etc/php-fpm.d/docker.conf

### DIFF
--- a/src/_base/harness/config/functions.yml
+++ b/src/_base/harness/config/functions.yml
@@ -85,6 +85,6 @@ function('version_compare', [version1, version2, operator]): |
   #!php
   $count1 = substr_count($version1, '.');
   $count2 = substr_count($version2, '.');
-  $version1 .= $count1 == 1 ? '.0' : '';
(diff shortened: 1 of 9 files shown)


kierenevans (dismissed):
Nice!

src/_base/harness/config/functions.yml
```diff
@@ -80,3 +80,11 @@ function('php_fpm_exporter_scrape_url', [hostname, pools]): |
       )
     );
   = $text;
+
+function('version_compare', [version1, version2, operator]): |
+  #!php
+  $count1 = substr_count($version1, '.');
+  $count2 = substr_count($version2, '.');
+  $version1 .= $count1 == 1 ? '.0' : '';
+  $version2 .= $count2 == 1 ? '.0' : '';
```
andytson: seems what you need to balance is the same number of components in each, not specifically more than one component

kierenevans (approved):

PR #301, opened 2020-05-21
Allow automated builds to show us errors

kierenevans:
When in static mode such as in Jenkins or in init/migrate steps on a kubernetes cluster, we need to be able to see the error messages of commands running via `run` rather than just `passthru`.


rgpjones (approved):
This change makes sense to me

PR #303, opened 2020-05-21
Resource Prefix support for Magento 1 & 2

kierenevans:
* Move redis and elasticsearch host/port environment variables to base harness from spryker
* Update magento 1 local.xml to use twig values
* Update magento 2 env.php to use environment variables

Should fix a resourcePrefix setting causing an init failure where `redis` is attempted to be connected to instead of `magento2-redis`


rgpjones (approved):
I think this looks okay. I get what's going on, anyway.

PR #304, opened 2020-05-25
Configure Magento 2 Cron

kierenevans:
* Configure Magento 2 Cron, but not in the development environment as it will slow down machines
* Fix bug that stops the environment starting when cron is in the services list
* Build and publish cron docker image if cron is in the services list
* Fix cron environment variables being treated as one string


src/_base/harness/scripts/enable.sh.twig
```diff
@@ -43,22 +43,37 @@ dynamic()
         passthru "docker-compose config --services | grep -v php-fpm | xargs docker-compose pull"
         passthru "docker-compose config --services | grep -v cron | xargs docker-compose build --pull"
         passthru docker-compose build cron
+        # Bring up all but cron
+        passthru "docker-compose config --services | grep -v cron | xargs docker-compose up -d"
```
tkotosz, line 47: noice!

src/_base/harness/scripts/enable.sh.twig
```diff
@@ -43,22 +43,37 @@ dynamic()
...
+    {% else %}
+        # Bring up all services
+        passthru docker-compose up -d
+    {% endif %}
+
     passthru docker-compose exec -T -u build console app init
+    # Bring up all services
+    passthru docker-compose up -d
```
tkotosz, line 76: hmm is this needed here?

tkotosz (approved):
left a question but other than that it looks good to me

src/_base/harness/scripts/enable.sh.twig
kierenevans: Brings up the cron which didn't get started earlier

tkotosz: ah I see...

PR #307, opened 2020-05-25
Move RabbitMQ to be available for Magento 2

kierenevans:
* Add RabbitMQ to base harness, moved from spryker harness
* Upgrade RabbitMQ to 3.8
* Add persistent storage for the helm chart for rabbitmq
* Configure Magento to use RabbitMQ if present in the services array.


tkotosz (approved):
nice work!

PR #309, opened 2020-06-01
Add Varnish

### LINKED ISSUE #14: Add option for varnish container as the entrypoint to the site in Magento 2.x harness

kierenevans:
If varnish is in the list of services:
* Use `varnish` as the entrypoint from traefik.
* Use `tls-offload` as the target for acceptance testing (behat) within the docker-compose environment to provide TLS support.
* Deploy `varnish` statefulset to pipeline environments as the entrypoint from the ingress.
* Deploy a configmap containing the varnish configuration, to avoid wiping out the cache on each deployment if built with docker image+configuration per-commit.
* Configure the "timeout chain" of varnish and nginx to be the max execution time of php plus one or two seconds. The idea is to never show a 502/504 when php is still working, though this may not be true if PHP is performing slow network requests as they block and don't count towards max_execution_time, or if the queue to access php-fpm from nginx is high.
* BREAKING CHANGE: `.Values.ingress` renamed to `.Values.ingress.type` to allow for more ingress configurations.
* Magento 2 VCL from https://github.com/magento/magento2/blob/2.3.5/app/code/Magento/PageCache/etc/varnish6.vcl used as the configuration.
* Magento 2 env.php configured for http_cache_hosts

Fixes https://github.com/inviqa/harness-base-php/issues/14


src/magento2/harness.yml
```diff
@@ -130,6 +133,9 @@ attributes:
         - run "magento config:set 'web/secure/use_in_frontend'  '1'"
         - run "magento config:set 'web/secure/use_in_adminhtml' '1'"
         - run "magento config:set 'web/seo/use_rewrites'        '1'"
+        - if [[ "$HAS_VARNISH" == "true" ]]; then
+            run "magento config:set 'system/full_page_cache/caching_application' '2'";
```
tkotosz, line 140: isn't there a few other config to set as well in this case? (like host port maybe?)

tkotosz (dismissed):
generally looks good to me. @dcole-inviqa could you please have a look at this, just to be on the safe side since it is a breaking change?

src/magento2/harness.yml
kierenevans: Nope, I think they are just for the VCL generator

tkotosz: ah I see :+1:

coledave (approved):

PR #310, opened 2020-06-02
Improve speed of permissions fixes

kierenevans:
`chown` adjusts every directory/file every time via `fchmodat` after first discovering them by scanning directories.

`find` user/group/perm discovers directories/files, filters out the already-correct directories/files and then changes the remaining files only.

Managed to save 2-4 minutes on a magento2 build with console/php-fpm/cron.


tkotosz (approved):
yay! thanks :bow:

PR #315, opened 2020-06-09
Enable local mail by default

kierenevans:
Moves the excellent work done for the spryker harness in https://github.com/inviqa/harness-base-php/pull/94 to be available to all frameworks.

Uses a NGINX TCP proxy to forward from the private network (i.e. console and php-fpm) to the shared network (mailhog)

### DIFF
--- a/src/_base/_twig/docker-compose.yml/application.yml.twig
+++ b/src/_base/_twig/docker-compose.yml/application.yml.twig
@@ -30,7 +30,6 @@
 {% include blocks ~ 'environment.yml.twig' %}
     networks:
       - private
-      - shared
 
   nginx:
 {% if @('app.build') == 'dynamic' %}
@@ -75,7 +74,6 @@
       - traefik.enable=false
     networks:
       - private
-      - shared
     environment:
 {{ deep_merge_to_yaml([@('services.php-base.environment'), @('services.php-fpm.environment')], 2, 6) | raw }}
 {% include blocks ~ 'environment.yml.twig' %}
@@ -83,3 +81,13 @@
 {% for pool in @('php-fpm.pools') %}
       - {{ pool.port }}
 {% endfor %}
+
+  relay:
+    build: .my127ws/docker/image/relay
+    labels:
+      - traefik.enable=false
+    networks:
+      private:
+        aliases:
+          - mailhog-relay
+      shared: {}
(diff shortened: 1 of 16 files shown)


src/_base/_twig/docker-compose.yml/application.yml.twig
```diff
@@ -74,6 +75,7 @@
       - traefik.enable=false
     networks:
       - private
+      - shared
```
andytson-inviqa: does this make talking to things like mysql problematic like php-fpm for nginx without the links?
have I opened a can of worms? :)

andytson-inviqa: is there any simple low cost (e.g. go-based binary on scratch) tcp relay that could be used to solve this without having to recursively use more links?

kierenevans: Multiple projects started at once with this change applied seem to be talking to the right places for me from the outside via traefik.

Inside the `console` container of one leads to this though, as you expect:
```
$ getent hosts nginx
172.19.0.3      nginx
172.19.0.8      nginx
172.19.0.19     nginx
172.19.0.7      nginx
172.19.0.14     nginx
```

Hmmm

andytson-inviqa: my expectations is php -> mysql would be round-robining the mysql service IP addresses, but not clever enough to pick up a different server with failing auth on the incorrect ones. Same for any other service we might need php to talk to, so I don't think we should introduce the links to every service that php talks to.
a container that tcp relays can be on the private + shared network while php sticks to private

andytson-inviqa: oh, actually, it wouldn't, as mysql isn't on the shared network

andytson-inviqa: it is annoying for php (like console) to nginx tho, but that's a different issue

kierenevans: Now implemented as a nginx tcp proxy_pass between console/php-fpm talking to `mailhog-relay` in the private network and `mailhog` in the shared one.
How's that? :)

andytson-inviqa: I'd "resolved" the conversation as I realised it's unlikely a problem, due to other services generally not being on shared network.

If you think a solution still needed like you have done, then that's ok

andytson-inviqa: I suppose you mean it'll be a problem for php -> nginx

kierenevans: Yeah, best all around to not join the shared network

andytson-inviqa (approved):

PR #319, opened 2020-06-12
Speed up Spryker build

kierenevans:
* Use pipeline/production mode for static tests
* Turn off opcache validate_timestamps when in production mode
* Use shallow clone of the b2c demoshop repo
* Fix cronjobs not running correctly out of the box in production mode
* Remove need to run two composer updates by merging symfony/console as root requirement in `composer-harness.json`. Updates to latest of 3.4 in case of security fixes.
* Tune opcache for spryker

Production mode builds gets a fail whale on all Zed/Yves endpoints as https://github.com/spryker-shop/b2c-demo-shop/blob/201903.0-p1/src/Pyz/Zed/Application/ApplicationDependencyProvider.php#L76 is enabled even if `spryker/web-profiler` is not present - it's only present for require-dev.

This is fixed in a more recent version of the b2c demo shop, which I believe was on the radar for upgrading to.

This PR gives from 0 to 5 minute speedup each build, saving a maximum total of 15 minutes across the three suites.

### DIFF
--- a/src/spryker/docker/image/console/root/lib/task/skeleton/apply.sh.twig
+++ b/src/spryker/docker/image/console/root/lib/task/skeleton/apply.sh.twig
@@ -30,9 +30,7 @@ function merge_composer_json()
     run "mv $merged_file $core_file"
 
     local new_composer_packages=""
-    new_composer_packages="$(jq -r '.["require-dev"] | keys[]' $override_file | tr '\n' ' ')"
+    new_composer_packages="$(jq -r '.["require-dev"] * .["require"] | keys[]' $override_file | tr '\n' ' ')"
 
     [ -n "$new_composer_packages" ] && passthru "COMPOSER_PROCESS_TIMEOUT=600 composer update --with-dependencies --no-interaction --profile $new_composer_packages"
-
-    passthru "COMPOSER_PROCESS_TIMEOUT=600 composer require --no-interaction --profile symfony/console:3.4.30"
 }
--- a/src/spryker/docker/image/console/root/lib/task/skeleton/apply.sh.twig
+++ b/src/spryker/docker/image/console/root/lib/task/skeleton/apply.sh.twig
@@ -11,7 +11,7 @@ function task_skeleton_apply()
 
 function fetch_demoshop()
 {
-    run "git clone -b 201903.0-p1 {{ @('spryker.demoshop-url') }} /tmp/spryker"
+    run "git clone --depth 1 --branch 201903.0-p1 {{ @('spryker.demoshop-url') }} /tmp/spryker"
     run "shopt -s dotglob && mv /tmp/spryker/* /app"
     run "rm -rf /tmp/spryker"
     run "rm -rf /app/.git"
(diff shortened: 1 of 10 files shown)


src/spryker/docker/image/console/root/lib/task/skeleton/apply.sh.twig
```diff
@@ -29,9 +29,8 @@ function merge_composer_json()
     run "jq -s '.[0] * .[1]' $core_file $override_file > $merged_file"
     run "mv $merged_file $core_file"
 
-    local new_composer_packages="$(jq -r '.["require-dev"] | keys[]' $override_file | tr '\n' ' ')"
+    local new_composer_packages=""
+    new_composer_packages="$(jq -r '.["require-dev"] * .["require"] | keys[]' $override_file | tr '\n' ' ')"
```
hgajjar, line 33: why defining empty var first and then reassigning with different value?

src/spryker/docker/image/console/root/lib/task/spryker/init.sh
```diff
@@ -4,6 +4,5 @@ function task_spryker_init()
 {
     run "vendor/bin/console propel:install -o"
     passthru "vendor/bin/install -r docker -s queue-flush"
-    passthru "PGPASSWORD=$DB_PASS vendor/bin/install -r docker -s queue-worker"
```
hgajjar, line 7: If we remove the queue-worker from here, it would depend on async Jenkins process to process the queue messages.
In CI build, we are running acceptance tests right after init is finished, so there we don't have any way to check if Jenkins has finished the queue messages.
That's why we are manually processing all messages here using queue-worker command

src/spryker/harness.yml
```diff
@@ -28,6 +28,8 @@ attributes:
     mode: development
   php:
     version: 7.2
+    ini:
+      opcache.max_accelerated_files: 32531
```
hgajjar: Is this the number of files this spryker release contains? could we set it to a bit higher number like 40k?

kierenevans: Only 28867 files in the spryker install when finished with dev dependencies and 22570 without.
Value will be rounded up to nearest prime number in the list from https://www.php.net/manual/en/opcache.configuration.php#ini.opcache.max-accelerated-files
Next prime number after that is 65407

src/spryker/docker/image/console/root/lib/task/skeleton/apply.sh.twig
kierenevans: Shellcheck recommendation from https://github.com/koalaman/shellcheck/wiki/SC2155

src/spryker/docker/image/console/root/lib/task/spryker/init.sh
kierenevans: The `queue-flush` on line 6 is already clearing any messages that were in the queues that could be processed, based on https://github.com/inviqa/harness-base-php/blob/feature/speed-up-spryker-build/src/spryker/application/skeleton/config/install/docker.yml.twig#L90-L105

I might be missing something though

kierenevans: If running through the installer (so without a database dump), the demodata will populate several thousand messages, then process them in the next step: https://github.com/inviqa/harness-base-php/blob/feature/speed-up-spryker-build/src/spryker/application/skeleton/config/install/docker.yml.twig#L217-L234

hgajjar: Hmm, seems like the install/init steps are changed since I configured them at the beginning, we are now using dedicated install/init bash files instead of their twig versions.

You are right about your last comment, its running the queue-worker from install recipe, so no need to run it from init.sh. However, I'm wondering if we need queue-flush here as well.

src/spryker/harness.yml
hgajjar: oh, didn't know that 👍

kierenevans: 32531 is one of the prime numbers though, should we increase to 65407 to account for project files/extra deps?

src/spryker/docker/image/console/root/lib/task/spryker/init.sh
kierenevans: Removed :)

src/spryker/harness.yml
hgajjar: I have more than 35k php files in total in a spryker project, so yeah we should increase this number

hgajjar (approved):
Nice work 👍

PR #322, opened 2020-06-16
Fix simpletest usage for Drupal 8

kierenevans:
The container running the website (php-fpm) needs to be the container
running the simpletest phpunit run, as it does calls with basic auth
that let the sqlite database be used instead of MySQL.

Without this, drupal is detecting the lack of a sessions table in the php-fpm sqlite database (despite it present in the console sqlite database) and redirecting to the installer.

### DIFF
--- a/src/drupal8/application/overlay/Jenkinsfile.twig
+++ b/src/drupal8/application/overlay/Jenkinsfile.twig
@@ -1,0 +1,63 @@
+pipeline {
+    agent { label "my127ws" }
+    environment {
+        MY127WS_KEY = credentials('{{ @('jenkins.credentials.my127ws_key') }}')
+        MY127WS_ENV = "pipeline"
+    }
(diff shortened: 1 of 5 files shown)


src/drupal8/application/overlay/Jenkinsfile.twig
```diff
@@ -0,0 +1,63 @@
...
+            steps {
+                sh 'ws install'
+                sh 'ws exec composer install --dev --no-interaction'
+                milestone(10)
+            }
+        }
+        stage('Test')  {
+            parallel {
+                stage('quality')    { steps { sh 'ws exec composer test-quality'    } }
+                stage('unit')       { steps { sh 'ws test-unit'                     } }
```
kierenevans, line 19: This is the difference to _base version

coledave (approved):

PR #328, opened 2020-06-23
Avoid copying vendor folder to NGINX, apart from akeneo

kierenevans:
Speeds up the NGINX docker image build for at least magento2.

Allow projects to override array entries with `- ~` and not fail the build.

Hint from @tkotosz, with aim to speed up the console to nginx docker image build step.

### DIFF
--- a/src/_base/docker/image/nginx/Dockerfile.twig
+++ b/src/_base/docker/image/nginx/Dockerfile.twig
@@ -6,7 +6,7 @@ FROM nginx:1.17-alpine
 COPY root /
 
 {% if @('app.build') == 'static' %}
-{% for copy_directory in @('nginx.copy_directories') %}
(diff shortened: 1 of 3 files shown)


src/_base/docker/image/nginx/Dockerfile.twig
```diff
@@ -6,7 +6,7 @@ FROM nginx:1.17-alpine
 COPY root /
 
 {% if @('app.build') == 'static' %}
-{% for copy_directory in @('nginx.copy_directories') %}
+{% for copy_directory in @('nginx.copy_directories')|filter(v => v is not empty) %}
```
andytson-inviqa, line 9: would be nice if workspace was fixed to avoid this

andytson-inviqa (approved):

PR #329, opened 2020-06-23
Avoid copying node_modules to php-fpm

kierenevans:
Hint from @tkotosz, with aim to speed up the build for the php-fpm image from console image step.


andytson-inviqa:
I'm wondering why they need to persist in the console static image in the first place

tkotosz (approved):

kierenevans:
> I'm wondering why they need to persist in the console static image in the first place

Maybe we don't need to have them as the console image is for init/migrate steps and ad-hoc commands.

We'd need to remove them immediately after the FE build completes to avoid being committed to a layer.

I'll raise a separate PR exploring this :)

andytson-inviqa:
adhoc commands wont share the artifacts from console into the nginx container though

andytson-inviqa:
I suppose because we use static builds for CI, if we wanted to do node tests at runtime then that'd be relevant (though I don't like the idea)

kierenevans:
We have a step to install dev dependencies for composer into the running console container, so we can repeat for npm deps too

PR #331, opened 2020-06-24
Do not install development dependencies in built docker image

kierenevans:
Avoid installing development dependencies in the console docker image, which would then increase the time taken to get copied to the php-fpm docker image, and time to push to the docker registry.

For magento2, there may be development extensions that get brought in with composer when installing the development tools at runtime, so need to run setup:upgrade and recompile the DI container.

For drupal8, we need to get phpunit installed into the php-fpm container for simpletest to work, but that means installing composer in the container or copying it across from the console container after installing. Decided to go with copying from console.

There is also a fix for checking if the containers are running for static mode, which has been broken since https://github.com/inviqa/harness-base-php/commit/0179b92768e8095c845eb1bd55cfad28bc084c12#diff-098f6bcd4621d373cade4e832627b4f6R20


test
```diff
@@ -83,9 +88,13 @@ function run_pipeline_tests()
     # Check if we have a composer.json so we can run tests
     # Check inside the console container as static builds do not
     # update the local filesystem
-    if ws exec stat composer.json > /dev/null 2>&1; then
+    if ws exec stat /app/composer.json; then
```
andytson-inviqa: why is the silencing of output removed?

kierenevans: Was for debugging as it wasn't firing the tests every time. Fixed :+1:

kierenevans:
Build failing due to:
`■■■ > composer install --no-interaction --no-dev --optimize-autoloader` printing out the composer help.

I guess we need #351 merged in first for the new passthru work.

src/spryker/application/skeleton/composer-harness.json
```diff
@@ -30,9 +30,12 @@
       "@test-unit",
       "@test-acceptance"
     ],
-    "test-quality": [
+    "devtools-setup": [
```
andytson-inviqa: would `generate` be a better name?

andytson-inviqa: (similiar in terminology as `go generate`)

kierenevans: Updated :+1:

andytson-inviqa (approved):

PR #333, opened 2020-06-30
Fix varnish cache clearing for magento2 when using resource prefixes

### LINKED ISSUE #332: varnish hostname needs to be set dynamically
when deploying to a cluster with a resourcePrefix set the hostname for varnish will change accordingly, eg. `varnish-0.varnish-headless` might become `magento-varnish-0.magento-varnish-headless`.

currently this will stop magento from being able to purge cache entries.

kierenevans:
Fixes #332


coledave (approved):
nice fix, lgtm !

PR #338, opened 2020-07-15
Use docker-compose build/push and service push config

andytson-inviqa:
So that services can be added or removed easily

Due to intermediate image needs for build, this can't be easily resolved automatically without stronger language support (like php composer dependencies)

### DIFF
--- a/src/_base/harness/attributes/docker-base.yml
+++ b/src/_base/harness/attributes/docker-base.yml
@@ -64,7 +64,7 @@ attributes:
         POSTGRES_USER: = @('database.user')
         PGDATA: /var/lib/postgresql/data/pgdata
     cron:
-      publish: "= ('cron' in @('app.services') ? 'true' : 'false')"
+      publish: "= 'cron' in @('app.services')"
       environment: []
     varnish:
       environment:
--- a/src/_base/harness/attributes/docker-base.yml
+++ b/src/_base/harness/attributes/docker-base.yml
@@ -32,6 +32,7 @@ attributes:
       environment:
         FPM_HOST: php-fpm
     console:
+      enabled: true
       publish: true
       environment:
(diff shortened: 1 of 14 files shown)


src/_base/harness/attributes/docker-base.yml
```diff
@@ -26,17 +26,19 @@ attributes:
...
         VARNISH_HOSTNAME_TEMPLATE: "varnish-%d.varnish-headless"
     nginx:
+      publish: true
       ingress:
         annotations: {}
       environment:
         FPM_HOST: php-fpm
     console:
-      enabled: true
+      publish: true
```
andytson-inviqa, line 36:
```suggestion
      enabled: true
      publish: true
```

src/_base/harness/config/pipeline.yml
```diff
@@ -5,49 +5,28 @@ command('app build'):
...
-
-command('app build tls-offload'): |
-  #!bash(harness:/docker/image/tls-offload)|@
-  passthru docker build -t @('docker.repository'):@('app.version')-tls-offload .
-
-command('app publish'):
+command('app build <service>'):
   env:
-    HAS_CRON: "= ('cron' in @('app.services') ? 'true' : 'false')"
+    SERVICE:    = input.argument('message')
```
andytson-inviqa:
```suggestion
    SERVICE:    = input.argument('service')
```

kierenevans:
Does `docker-compose down --rmi local` clean up tagged images still? Seen a few things left around on Jenkins runners.

kierenevans:
Never mind, we are doing this separately in destroy step: https://github.com/inviqa/harness-base-php/blob/0.9.x/src/_base/harness/scripts/destroy.sh#L13-L16

kierenevans (approved):

PR #345, opened 2020-07-29
Deduce the docker registry_url from repository

andytson-inviqa:
Docker login doesn't work when the repository is using the default daemon registry (common when pushing to docker hub)

Also makes it clear there is no per-repository login credential store, only per-registry, so not good for parallel builds of different projects


kierenevans (approved):

kierenevans (changes requested):
Does the docker logout need changing too?

kierenevans (approved):

PR #347, opened 2020-07-29
Remove node_modules from console image. Clear cache

kierenevans:
Clean up after the installation for the static build.

### DIFF
--- a/src/_base/harness/attributes/common.yml
+++ b/src/_base/harness/attributes/common.yml
@@ -117,15 +117,20 @@ attributes.default:
     build:
       when: -f "package.json"
       steps:
-        - run "yarn install"
+        - |
+          if [ "$APP_BUILD" == "static" ] && [ -f package-lock.json ]; then
+            run "npm clean-install"
+          else
+            run "npm install"
+          fi
         - |
           if jq ".scripts.build != null" < package.json; then
-            run "yarn run build"
+            run "npm run build"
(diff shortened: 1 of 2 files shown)


src/_base/harness/attributes/common.yml
```diff
@@ -118,6 +118,11 @@ attributes.default:
       when: -f "package.json"
       steps:
         - run "npm install"
+        - |
+          if [ "$APP_BUILD" == "static" ]; then
+            run "rm -rf node_modules/"
+            run "npm cache clean --force"
+          fi
```
andytson-inviqa, line 134: hmm, do we need a assumption that this is changed or should we add a `passthru "npm build"` assumption?

kierenevans: Yeah was wondering if `npm run build` should be a convention, though this will cause it to fail if not present in package.json

andytson-inviqa: I think yarn is more convention as well

kierenevans: Updated to use yarn :+1:

kierenevans: I did switch to yarn, but reading our engineering guidelines we favour npm. I'll switch back to npm.

andytson-inviqa (approved):

PR #354, opened 2020-08-11
Pull external images in parallel before building service images

andytson-inviqa:


src/_base/harness/config/functions.yml
```diff
@@ -114,6 +114,30 @@ function('get_docker_external_networks'): |
   }
   = join(" ", $externalNetworks);
 
+function('docker_service_images'): |
```
kierenevans, line 117: I think we need to account for the `build: path/to/context` too

andytson-inviqa: no, `docker-compose config` converts it to build.context

kierenevans: Sweet :)

kierenevans (approved):

PR #359, opened 2020-08-14
Replace global zed host with store specific hosts

hgajjar:
Spryker Yves should connect to the Zed of same store, for example Yves running in DE store context must connect to Zed with DE store context only.

So I have split the common zed host into store specific ones, and updated the spryker's config_local.php to use the related Zed host from Yves.


andytson-inviqa:
I'm a bit worried about how projects are using this currently compared to this change

hgajjar:
> I'm a bit worried about how projects are using this currently compared to this change

True, but unfortunately this is how default Spryker works. In its core design, they expect that the backoffice (zed) of each store would be a separate domain, which would be managed by different teams. There is no concept of "common" zed for all stores.

andytson-inviqa:
I think it's right to do this, but currently it's not backwards compatible, so semver we might not want to release it for 0.10.x

If it can be made backwards compatible then a major release can remove the old way later

hgajjar:
> I think it's right to do this, but currently it's not backwards compatible, so semver we might not want to release it for 0.10.x
> 
> If it can be made backwards compatible then a major release can remove the old way later

ok, I will check if I can make it backward compatible

andytson-inviqa (approved):

PR #364, opened 2020-08-18
Fix not having output for failed commands

kierenevans:
The `bin/app` `set -e` was not allowing the separated `bash` call to fail and still continue onwards to show the /tmp/my* files.

### DIFF
--- a/src/_base/docker/image/console/root/lib/sidekick.sh
+++ b/src/_base/docker/image/console/root/lib/sidekick.sh
@@ -52,7 +52,7 @@ run()
         prompt
         if [ "${DEPRECATED_MODE}" = "yes" ]; then
             echo "  > ${COMMAND_DEPRECATED[*]}" >&2
-            COMMAND=(bash -c "${COMMAND_DEPRECATED[@]}")
+            COMMAND=(bash -e -c "${COMMAND_DEPRECATED[@]}")
         else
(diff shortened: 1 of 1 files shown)


src/_base/docker/image/console/root/lib/sidekick.sh
```diff
@@ -52,18 +52,17 @@ run()
...
         if [ "${DEPRECATED_MODE}" = "yes" ]; then
             echo "  > ${COMMAND_DEPRECATED[*]}" >&2
-            setCommandIndicator "${INDICATOR_RUNNING}"
-            bash -c "${COMMAND_DEPRECATED[@]}" > /tmp/my127ws-stdout.txt 2> /tmp/my127ws-stderr.txt
+            COMMAND=(bash -e -c "${COMMAND_DEPRECATED[@]}")
         else
-            echo "  >$(printf ' %q' "${COMMAND[@]}")" >&2
-            setCommandIndicator "${INDICATOR_RUNNING}"
-            "${COMMAND[@]}" > /tmp/my127ws-stdout.txt 2> /tmp/my127ws-stderr.txt
+          echo "  >$(printf ' %q' "${COMMAND[@]}")" >&2
```
andytson-inviqa:
```suggestion
            echo "  >$(printf ' %q' "${COMMAND[@]}")" >&2
```

andytson-inviqa (approved):

PR #366, opened 2020-08-18
Fix drush relative docroot fails

kierenevans:
Encountered ./docroot being used internally for `ws drush updb -y`, which fails. Use absolute paths to avoid this.


mazeikis (approved):
👍

PR #373, opened 2020-08-20
Make the Spryker jenkins runner single executor, multi containers

andytson-inviqa:
It's much easier to ensure memory limits per container than per executor

Current settings and default memory limits cause OOM errors


jameshalsall (dismissed):

hgajjar (approved):
LGTM

PR #375, opened 2020-08-20
Switch back to the default docker stop signal in cli images

andytson-inviqa:
Most cli processes don't trap SIGQUIT, so end up being SIGKILLed after 10 seconds of waiting with no graceful handling

### DIFF
--- a/src/_base/docker/image/console/Dockerfile.twig
+++ b/src/_base/docker/image/console/Dockerfile.twig
@@ -1,4 +1,6 @@
 FROM {{ @('docker.image.console') }}
+# fix upstream signal
+STOPSIGNAL SIGTERM
 
 COPY .my127ws/docker/image/console/root /
 RUN chown -R build:build /home/build \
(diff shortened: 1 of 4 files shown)


hgajjar (approved):
LGTM, but I'm not sure if this could have any other side-effects, would be nice to have second opinion

kierenevans (approved):
SIGTERM is the default, and SIGQUIT is needed only for fpm in https://github.com/docker-library/php/blob/2e2df0625bdf4e882dfcdb314970424abb24f36f/fpm-Dockerfile-block-2#L43.
If we aren't running fpm, it's good to set back to the default :+1:

kierenevans:
Oh perhaps this should be in the https://github.com/my127/docker-php

andytson-inviqa:
>Oh perhaps this should be in the https://github.com/my127/docker-php

In any case the cron service is using the harness-built php-fpm image as a intermediate image, so is needed anyway. That and any upstream image can be specified in the attributes as well, so I think it best in both places

PR #377, opened 2020-08-21
Make jenkins service optional for Spryker harness

hgajjar:
As the jenkins service requires lots of resources, it's not ideal to use it in pipeline and local environments as it is. There is some thought to move jenkins master to a shared service in pipeline env, so that can be shared across all pipeline environments, but that might require quite some work.

So for now I have make this service optional and disabled it by default, so we can already add a demo k8s env for it with argocd for example.

Adding cron service support to Spryker harness might be useful, but it would have problems with limiting concurrent crons leading to resource issues as Andy raised.


andytson-inviqa:
>but it would have problems with limiting concurrent crons leading to resource issues as Andy raised.

As well as single point of failure if used in critical environments, and not make good use of the full cluster as one container

For dev experience locally and CI it's good though (not sure whether the CI automated testing environment is around long enough to need the crons running though)

src/spryker/harness.yml
```diff
@@ -20,8 +20,6 @@ attributes:
       - redis
       - elasticsearch
       - rabbitmq
-      - jenkins
-      - jenkins-runner
```
andytson-inviqa: As we're looking to keep backwards compatibility can it be on by default?

Also atm this service config is required for docker images to be pushed and deployed, we need to have a different way of saying it shouldn't be used for dev/CI(built but not started) but should for deployment. This could be another PR though if backwards compatibility is maintained here

hgajjar:
> > but it would have problems with limiting concurrent crons leading to resource issues as Andy raised.
> 
> As well as single point of failure if used in critical environments, and not make good use of the full cluster as one container
> 
> For dev experience locally and CI it's good though (not sure whether the CI automated testing environment is around long enough to need the crons running though)

CI env does not need crons running, as we are running the rabbitmq sync command at the end of init. However, the pipeline environments (QA/UAT/etc) would need it

kierenevans:
I've rebased this on top of latest 0.10.x with #382 merged in, applied #382 to the new enable.sh.twig location for jenkins-runner and allowed cron and jenkins-runner to be in the same environment.

src/spryker/harness.yml
```diff
@@ -20,8 +20,6 @@ attributes:
       - redis
       - elasticsearch
       - rabbitmq
```
kierenevans: To have the backwards compatibility described in the comments above:
```suggestion
      - rabbitmq
      - jenkins
      - jenkins-runner
```

kierenevans (approved):

PR #378, opened 2020-08-21
Allow easy publish environment settings in Jenkins

kierenevans:
Stops us needing to override the Jenkinsfile template on every project

### DIFF
--- a/src/_base/application/overlay/Jenkinsfile.twig
+++ b/src/_base/application/overlay/Jenkinsfile.twig
@@ -29,11 +29,13 @@ pipeline {
         }
 {% if @('pipeline.publish.enabled') == 'yes' %}
         stage('Publish') {
+{% if @('pipeline.publish.environment') %}
             environment {
 {% for key, value in @('pipeline.publish.environment') %}
-                {{ key }} = {{ value|raw }}
+                {{ key }} = {{ value }}
 {% endfor %}
             }
+{% endif %}
             when { not { triggeredBy 'TimerTrigger' } }
             steps {
                 milestone(50)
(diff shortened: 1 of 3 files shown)


src/_base/harness/attributes/common.yml
```diff
@@ -38,6 +38,9 @@ attributes.default:
     publish:
       enabled: no
       services: = publishable_services(@('services'))
+      # For defining environment variables in Jenkins, e.g. loading up docker username/password from a Jenkins
+      # credential
+      environment: ~
```
andytson-inviqa: does this need to be {} for the default value to be iterated 0 times instead of erroring?

src/_base/application/overlay/Jenkinsfile.twig
```diff
@@ -29,6 +29,11 @@ pipeline {
         }
 {% if @('pipeline.publish.enabled') == 'yes' %}
         stage('Publish') {
+            environment {
+{% for key, value in @('pipeline.publish.environment') %}
+                {{ key }} = {{ value|raw }}
```
andytson-inviqa: wondering if we should skip raw now as workspace 0.1.3 (which turns off autoescape) already depended on in other places for a while

andytson-inviqa (dismissed):

src/_base/application/overlay/Jenkinsfile.twig
kierenevans: raw removed :+1:

src/_base/harness/attributes/common.yml
kierenevans: You reminded me I had issues with empy `environment {}` being a groovy error. Updated the default values to be `{}` and checks in the Jenkinsfile template.

andytson-inviqa (approved):

PR #383, opened 2020-08-25
Image pull config must be present before app-init job

kierenevans:
Docs: https://argoproj.github.io/argo-cd/user-guide/sync-waves/

### DIFF
--- a/src/_base/helm/app/templates/application/image-pull-config.yaml
+++ b/src/_base/helm/app/templates/application/image-pull-config.yaml
@@ -9,8 +9,9 @@ kind: Secret
 metadata:
   name: {{ .Values.resourcePrefix }}image-pull-config
   annotations:
-    sealedsecrets.bitnami.com/cluster-wide: "true"
+    argocd.argoproj.io/sync-wave: "4"
 {{ if .Values.feature.sealed_secrets }}
+    sealedsecrets.bitnami.com/cluster-wide: "true"
 spec:
   template:
     type: kubernetes.io/dockerconfigjson


coledave (approved):

src/_base/helm/app/templates/application/image-pull-config.yaml
```diff
@@ -9,8 +9,9 @@ kind: Secret
 metadata:
   name: {{ .Values.resourcePrefix }}image-pull-config
   annotations:
-    sealedsecrets.bitnami.com/cluster-wide: "true"
+    argocd.argoproj.io/sync-wave: "4"
```
andytson-inviqa, line 12: documentation says default is 0 and lower number waves are done before higher, so this would already be the case, right, unless using nevative number waves?

it's useful for everything that has no dependency on other resources to be in the same wave so as to be more parallel for speed, so I'm not sure but the default might be better?

andytson-inviqa: what you might have been finding is the secret wasn't decrypted into a secret resource in time in a real run maybe? Increasing the wave to make it happen later only increases that chance

kierenevans: For whatever reason, one of our environments doing this conversion did not put the sealed secret for image pull config on the cluster before trying to do app-init.

It's true that the default of 0 should have been sufficient, but let's be explicit about the ordering.

andytson-inviqa: is every resource like this then? should we set it to 0 for explicitness that there's no dependency?

kierenevans: If and when we spot a problem, can change it :+1:

andytson-inviqa: likewise app-init will have a need for the environment secrets

PR #385, opened 2020-08-26
Recreate services that use persistent volumes to avoid multi-node usage attempts

kierenevans:
Read Write Once volumes can't be used on multiple nodes in digitalocean kubernetes or google kubernetes engine.

Any deployment spec changes for services such as mysql fail to launch when using rolling update with a RWO persistent volume attached and manual intervention is needed to scale down the service, then back up.

### DIFF
--- a/src/_base/helm/app/templates/service/elasticsearch/deployment.yaml
+++ b/src/_base/helm/app/templates/service/elasticsearch/deployment.yaml
@@ -10,6 +10,10 @@ spec:
   selector:
     matchLabels:
       app.service: {{ .Values.resourcePrefix }}elasticsearch
+{{- if .Values.persistence.elasticsearch.enabled }}
+  strategy:
+    type: Recreate
+{{- end }}
   template:
     metadata:
       labels:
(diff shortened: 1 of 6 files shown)


andytson-inviqa (changes requested):
these services support configuring readwritemany in the persistence configuration, so shouldn't need this if they are using that mode

`.Values.persistence.elasticsearch.accessMode`

kierenevans:
@andytson-inviqa Do any of these services support reading from and writing to the same datastore from two instances at once?

* MySQL doesn't start up if another mysqld process is actively using the datastore.
* Redis would be warring as it's doing a dump to disk from memory, where it overwrites the dump file.
* RabbitMQ definitely hates it.
* Postgres is the only one I'm unsure of.

If we want to do clustering/ha, we can use a statefulset with a volumeClaimTemplate to get distinct volumes for the data storage.

andytson-inviqa:
ok can you change the title of the PR then to reflect the reason being persistent volumes rather than specifically RWO?

andytson-inviqa (approved):

andytson-inviqa:
or maybe we remove support for changing accessMode for them instead?

PR #389, opened 2020-09-01
Fix php syntax check not exiting with non-zero exit code on error

kierenevans:
Also expand checks for drupal8 to cover the same PHP extensions as phpcs.


tkotosz (dismissed):
LGTM

tkotosz (approved):

PR #391, opened 2020-09-07
MISC Publish images and chart only when needed

andytson-inviqa:
* for backwards compatibility, all branches are still published, however that will be changed in a major release
* if preview enabled, then PRs that target the default branch (if pipeline.publish.branches is limited to specific branches)
* rename git.main_branch to git.default_branch as main has meanings that a develop branch process is confusing with (especially if renaming master to main)

### DIFF
--- a/src/_base/application/overlay/Jenkinsfile.twig
+++ b/src/_base/application/overlay/Jenkinsfile.twig
@@ -4,7 +4,7 @@ pipeline {
         MY127WS_KEY = credentials('{{ @('jenkins.credentials.my127ws_key') }}')
         MY127WS_ENV = "pipeline"
     }
-    triggers { cron(env.BRANCH_NAME == '{{ @('git.main_branch') }}' ? 'H H(0-6) * * *' : '') }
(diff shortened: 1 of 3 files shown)


kierenevans:
The ArgoCD based environments don't use preview or qa "environments" as defined in the harness.

Only `pipeline.publish`.

andytson-inviqa:
> The ArgoCD based environments don't use preview or qa "environments" as defined in the harness.
> 
> Only `pipeline.publish`.

a) backwards compatibility with qa and existing preview usage. If I didn't put the qa bit in, existing old process qa deployments wouldn't have any docker images to deploy
b) currently it is publishing every single branch to the docker and git cluster repository, we need a way to stop that also avoid docker publish when not needed

therefore if we do introduce preview environments to it, we need to say which should be published, and we do that by using pull requests rather than wasting  preview in environment on WIP

most projects I've put into harness I've always added the `branch 'master'` to avoid (b)

kierenevans (approved):

PR #394, opened 2020-09-08
Configure symfony harness

kierenevans:
TODO:
- [x] Database support
- [x] Installation steps


src/symfony/application/skeleton/composer.json.twig
```diff
@@ -0,0 +1,150 @@
...
+        "symfony/security-bundle": "5.1.*",
+        "symfony/serializer": "5.1.*",
+        "symfony/string": "5.1.*",
+        "symfony/translation": "5.1.*",
+        "symfony/twig-bundle": "5.1.*",
+        "symfony/validator": "5.1.*",
+        "symfony/web-link": "5.1.*",
+        "symfony/yaml": "5.1.*",
+        "twig/extra-bundle": "^2.12|^3.0",
+        "twig/twig": "^2.12|^3.0"
```
andytson-inviqa: would it be worth only going for the minumum that symfony/flex sets up?

kierenevans: Looks like the sample project we have chose v3.0.5 in https://github.com/inviqa/symfony-sample/blob/main/composer.lock#L6012, so I can adjust this to be the ^3.0 requirement?

andytson-inviqa: sorry I meant the whole dependency list, using only what symfony/flex would initially install

kierenevans: This is the composer.json (with additional dev dependencies added by me) generated by `composer create-project symfony/website-skeleton test-symfony`

We could use the "symfony/skeleton" but then won't have a website to test, or we can run the command upon applying the skeleton to a project like the spryker harness does, with a composer.json merge to add our development tools? Or something else?

Followed https://symfony.com/doc/current/setup.html#creating-symfony-applications

andytson-inviqa: symfony/skeleton includes the web bits as well, by default just hosts the welcome page (prob same as with website-skeleton, just with less packages installed)

kierenevans: Okay I'll re-generate it with just skeleton instead of website-skeleton

kierenevans: Normal skeleton ends up with the following and no doctrine:
```json
    "require": {
        "php": ">=7.2.5",
        "ext-ctype": "*",
        "ext-iconv": "*",
        "symfony/console": "5.1.*",
        "symfony/dotenv": "5.1.*",
        "symfony/flex": "^1.3.1",
        "symfony/framework-bundle": "5.1.*",
        "symfony/yaml": "5.1.*"
    },
    "require-dev": {
    },
```

What do we want to go with as a "better practise application"?

andytson-inviqa: considering not all apps use doctrine and it's a single command to install and configure with flex this can be documentation. Symfony devs should be aware of this and this allows minimal apps OTB

src/symfony/harness/attributes/docker.yml
```diff
@@ -0,0 +1,9 @@
+attributes:
+  services:
+    php-base:
+      environment:
+        APP_DEBUG: "0"
+        APP_ENV: "= (@('app.mode') == 'development' ? 'dev' : 'prod')"
+      environment_secrets:
+        APP_SECRET: = @('symfony.secret')
```
andytson-inviqa: I think it might be a bit confusing having a `symfony.secret` that is only for local environment and might lead to reuse

src/symfony/harness/attributes/docker.yml
```diff
@@ -0,0 +1,9 @@
...
+        DATABASE_URL: = 'mysql://' ~ @('database.user')  ~ ':' ~ @('database.pass')  ~ '@' ~ @('database.host') ~ ':' ~ @('database.port')  ~ '/' ~ @('database.name')  ~ '?serverVersion=' ~ @('database.platform_version')
```
andytson-inviqa: could be worth a skeleton config/...yml with the following, so only the password needs to be encrypted, and same env vars work with asset apply
```
parameters:
  env(DATABASE_URL): "%env(DATABASE_PLATFORM)%://%env(DATABASE_USER)%:%env(DATABASE_PASSWORD)%@%env(DATABASE_HOST)%:%env(DATABASE_PORT)%/%env(DATABASE_NAME)%?serverVersion=%env(DATABASE_VERSION)%

kierenevans: Great idea. Done!

One caveat is that .env can't have DATABASE_URL in, as it takes precedence over the config/*/*.yaml

src/symfony/harness/attributes/docker.yml
kierenevans: Removed

andytson-inviqa (approved):

PR #405, opened 2020-09-18
MISC Fix testing notes in readme

tkotosz:
```
$ echo 'attribute(\'mutagen\'): no' >> workspace.override.yml
bash: syntax error near unexpected token `)')
```

vs
```
$ echo "attribute('mutagen'): no" >> workspace.override.yml
$
```


kierenevans (approved):

andytson-inviqa (approved):
thanks, that was bugging me as well

PR #416, opened 2020-09-22
Set error reporting to E_ALL for PHP CLI

dantleech:


andytson-inviqa:
I'll do it the way I said in the chat

PR #417, opened 2020-09-22
Move php error reporting so it is available for both cli and fpm

andytson-inviqa:


src/_base/harness/attributes/common.yml
```diff
@@ -191,6 +189,8 @@ attributes.default:
         variables_order: GPCS
     ini:
       enable_dl: Off
+      error_reporting: "E_ALL"
+      log_errors: On
```
kierenevans: Does this send to stderr or to file?

andytson-inviqa: good point, I'll put it back in fpm

kierenevans (dismissed):

kierenevans:
I'm unsure what happens with the pipeline env's merge of php.fpm.ini.error_reporting in https://github.com/inviqa/harness-base-php/blob/0.10.x/src/_base/harness/attributes/environment/pipeline.yml#L15

I can investigate tomorrow if you don't have time.

andytson-inviqa:
I've checked php cli and it renders the last of the two in a php.ini. php-fpm's docker writes to the php.ini file both settings, php.ini first, php.fpm.ini last, so it'll get the fpm pipeline settings

It does make me wonder if pipeline cli should also be that?

kierenevans:
Yeah I think to avoid confusion due to duplicated entries (and E_ALL for console in pipeline) we should move:
https://github.com/inviqa/harness-base-php/blob/0.10.x/src/_base/harness/attributes/environment/pipeline.yml#L15
https://github.com/inviqa/harness-base-php/blob/0.10.x/src/spryker/harness/attributes/environment/pipeline.yml#L15
to be under `php.ini.error_reporting`, so it overrides the default set in common.yml cleanly.

kierenevans (approved):

PR #419, opened 2020-09-25
Use base OS appropriate Spryker jenkins runner java version

andytson-inviqa:
Only currently those supported by the debian distrubution
8 - stretch
11 - buster

### DIFF
--- a/src/spryker/harness.yml
+++ b/src/spryker/harness.yml
@@ -95,6 +95,9 @@ attributes:
     host: jenkins
     port: 8080
     external_host: = 'jenkins-' ~ @('hostname')
+  jenkins_runner:
+    # used by the Jenkins JNLP agent
+    java_version: 8
   yves:
     external_hosts:
       DE: = 'yves-de-' ~ @('hostname')
--- a/src/spryker/harness.yml
+++ b/src/spryker/harness.yml
@@ -95,9 +95,6 @@ attributes:
     host: jenkins
     port: 8080
     external_host: = 'jenkins-' ~ @('hostname')
-  jenkins_runner:
-    # used by the Jenkins JNLP agent
-    java_version: 8
   yves:
     external_hosts:
       DE: = 'yves-de-' ~ @('hostname')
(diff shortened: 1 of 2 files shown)


src/spryker/harness.yml
```diff
@@ -95,6 +95,9 @@ attributes:
     host: jenkins
     port: 8080
     external_host: = 'jenkins-' ~ @('hostname')
+  jenkins_runner:
+    # used by the Jenkins JNLP agent
+    java_version: 8
```
kierenevans: Can we switch this automatically like we do for the console/php-fpm image?

andytson-inviqa: if php version is 7.4 assume buster so assume 11? :)

kierenevans: Ah we have 7.3 buster images too, hmm

hgajjar: only latest spryker release supports php 7.4, which isn't supported by harness yet.

andytson-inviqa: yep, tho I've done this because I need to use the buster 7.3 image

andytson-inviqa: I'm not sure existing projects are held back in upgrading because of the harness tho (tho zed stuff still needs sorting out)

kierenevans (dismissed):

src/spryker/harness.yml
kierenevans: Could do an attribute for base OS, but too complicated if you can just set this on the project. Approved :)

kierenevans: Oh there is:
```
default-jre/oldstable 2:1.8-58+deb9u1 amd64
  Standard Java or Java compatible Runtime
```

This picks openjdk-8-jre on stretch and openjdk-11-jre on buster

kierenevans (approved):

src/spryker/harness.yml
andytson-inviqa: I would also wonder if default-jre-headless as well (no gui bits for the app)

kierenevans: Potentially, though I guess PDF rendering tasks will need some X components/fonts. Unsure if Spryker does any of that :)

PR #420, opened 2020-09-25
Copy redis-cli from the first enabled redis service's image

andytson-inviqa:


src/_base/docker/image/console/Dockerfile.twig
```diff
@@ -2,6 +2,11 @@ FROM {{ @('docker.image.console') }}
 # fix upstream signal
 STOPSIGNAL SIGTERM
 
+{% for service in [@('services.redis'), @('services.redis_session')] | map(v => v.enabled) | slice(0, 1) %}
+# copy the redis-cli from the first enabled redis service image
+COPY --from={{ service.image }} /usr/local/bin/redis-cli /usr/local/bin/redis-cli
+{% endfor %}
```
andytson-inviqa, line 8: I'm not sure on this, in that do we want to care what happens if a different image is used that doesn't have redis-cli in that location?

kierenevans: In that instance I think overriding this template on the project would be okay.

Alternatively we can do a FROM build stage where it does `mv "$(which redis-cli)" /usr/local/bin/redis-cli`

kierenevans (approved):

kierenevans (changes requested):
A few builds are failing on:
```bash
COPY --from= /usr/local/bin/redis-cli /usr/local/bin/redis-cli
```

andytson-inviqa:
all the ones that use redis

kierenevans (approved):

src/_base/docker/image/console/Dockerfile.twig
tkotosz: > I'm not sure on this, in that do we want to care what happens if a different image is used that doesn't have redis-cli in that location?

Also what if the other image has a different redis version in it (which is currently technically allowed)?

e.g. I ran into this here: https://github.com/inviqa/harness-base-php/pull/425
```
tkotosz ~/Sites/harness-base-php/tests/magento2-test (feature/magento-upgrade-to-240)
$ dc exec redis redis-cli --version
redis-cli 5.0.9
tkotosz ~/Sites/harness-base-php/tests/magento2-test (feature/magento-upgrade-to-240)
$ dc exec redis-session redis-cli --version
redis-cli 4.0.14
```

andytson-inviqa: I guess technically the higher version of the two is best, rather than the first enabled one, however that is far more of an edge case that doesn't have much in implications, and as you've shown, devs will already have both redis-cli available.

PR #426, opened 2020-09-26
Akeneo - fix random build failures

kierenevans:
* Wait for elasticsearch before install
* Bring up console first, so it can do it's entrypoint.dynamic.sh chown of /app to match the build user.
* Then bring up php-fpm and everything else, so www-data will be the same UID as the build user and be able to write to `var/logs/dev.log`.

Previously, UID 1000 owned the /app/var/ folder but www-data was UID 33 still, due to a parallel startup.

### DIFF
--- a/src/_base/harness/scripts/enable.sh.twig
+++ b/src/_base/harness/scripts/enable.sh.twig
@@ -50,6 +50,9 @@ dynamic()
         passthru docker-compose build jenkins-runner
     {% endif %}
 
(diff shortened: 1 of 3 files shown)


tkotosz (approved):
looks good, you might need to add the wait to the migrate steps as well but I don't know how akeneo works so :+1:

kierenevans:
doctrine migrations aren't talking to elasticsearch so should be okay 🤞

PR #427, opened 2020-09-28
Spryker - Do not wait 60 seconds if the queue is empty

kierenevans:
Now that we are on the January 2020 release of Spryker demoshop, we can use the `--stop-when-empty` flag to save 3 minutes on the build of Spryker per static/dynamic build, so 9 minutes!

Previously added to #319 but reverted due to older version of the demoshop being used.

### DIFF
--- a/src/spryker/application/skeleton/config/install/docker.yml.twig
+++ b/src/spryker/application/skeleton/config/install/docker.yml.twig
@@ -220,7 +220,7 @@ sections:
 
     queue-worker:
         run:
-            command: "vendor/bin/console queue:worker:start -vvv"
+            command: "vendor/bin/console queue:worker:start --stop-when-empty -vvv"
             stores:
                 - DE
                 - AT


src/spryker/application/skeleton/config/install/docker.yml.twig
```diff
@@ -220,7 +220,7 @@ sections:
 
     queue-worker:
         run:
-            command: "vendor/bin/console queue:worker:start -vvv"
+            command: "vendor/bin/console queue:worker:start --stop-when-empty -vvv"
```
hgajjar, line 223: This flag was already there, but there was an issue that it wasn't processing all pending messages (sometimes). Tho they might have fixed it in January release.

kierenevans: Something we can test with behat to save 9 minutes?

hgajjar: It's random, so can't really test!

But if you can make the CI build green, we can try to merge it and see how it goes. If it causes issue, can revert it. As it won't affect existing projects, it should be ok

kierenevans: Build isn't green due to an error deleting php docker images, but the spryker part is all okay now

hgajjar (approved):

PR #428, opened 2020-09-28
Test using 3 Jenkins workers

kierenevans:
Rather than a worker per framework, use 3 Jenkins workers.

Keeps less workers around for the timeout after the build is done (say on the nightly crons), but uses the assigned nodes more.

Builds take around the same time as Spryker is the the longest, but we get feedback for, say, magento 1's build a lot later.

Unsure if this is preferable!

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,6 +5,7 @@ pipeline {
     }
     options {
         buildDiscarder(logRotator(daysToKeepStr: '30'))
+        parallelsAlwaysFailFast()
     }
     triggers { cron(env.BRANCH_NAME ==~ /^\d+\.\d+\.x$/ ? 'H H(0-6) * * *' : '') }
     stages {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,7 +10,7 @@ pipeline {
     stages {
         stage('Build and Test') {
             parallel {
-                stage('1. PHP, Symfony, Wordpress, Akeneo') {
+                stage('1. PHP, Drupal 8, Akeneo') {
                     agent { label "my127ws" }
                     stages {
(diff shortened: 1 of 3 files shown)


andytson-inviqa:
for me it seems almost preferable to have spryker on it's own node, so slow compared to the rest

Jenkinsfile
```diff
@@ -5,46 +5,126 @@ pipeline {
...
-                    axis {
-                        name 'FRAMEWORK'
-                        values 'akeneo', 'drupal8', 'magento1', 'magento2', 'php', 'spryker', 'symfony', 'wordpress'
+        stage('Build and Test') {
+            parallel {
+                stage('1. PHP, Drupal 8, Akeneo') {
+                    agent { label "my127ws" }
+                    stages {
+                        stage('Prepare') {
+                            steps { sh './build' }
```
andytson-inviqa: build can't be in the parallel stage, as it needs to be available before the parallel stage, otherwise the tests can easily fail through race conditions

kierenevans: There's no agent selected until this stage, so nothing checked out. This is the first step on a worker post-checkout.

kierenevans:
> for me it seems almost preferable to have spryker on it's own node, so slow compared to the rest

Agreed, though it becomes more competitive with #427 added

Wordpress is there to try to fail fast if there's an obvious problem with the base harness.

Jenkinsfile
andytson-inviqa: so.. do you mean it pulls this stage out into a non-parallel stage automatically?

andytson-inviqa: oh, I see the only parallel bit is the stage('#. ....'), these are all sequential

PR #442, opened 2020-10-02
Fix resource prefixed magento and akeneo

kierenevans:
Would be something like `magento-elasticsearch:9200` on kubernetes


tkotosz (dismissed):

andytson-inviqa (dismissed):

kierenevans:
ELASTICSEARCH_HOST not passed through to app-init/app-migrate. Investigating

kierenevans:
Fixed in ff44637

andytson-inviqa (approved):

PR #443, opened 2020-10-02
Fix elasticsearch mountpoint

### LINKED ISSUE #441: [k8s] Elasticsearch does not boot with persistent storage enabled
Fatal error encountered in logs:
```
ElasticsearchException[failed to bind service]; nested: AccessDeniedException[/usr/share/elasticsearch/data/nodes];
```

This is due to the persistent storage directory being owned by root when created, so we need an initContainer to chown the empty folder like we do with webapp.

kierenevans:
Fixes #441


tkotosz (approved):

tkotosz:
note: based on https://kubernetes.io/docs/tasks/configure-pod-container/security-context/ (https://raw.githubusercontent.com/kubernetes/website/master/content/en/examples/pods/security/security-context.yaml)

PR #444, opened 2020-10-02
Configure pipeline environments to talk to mailhog for sendmail

kierenevans:
Currently the sendmail_path php.ini setting is compiled into the docker image.

We need a way to influence this on the kubernetes cluster to direct to a different service name than "mailhog-relay:1025"

This PR introduces a script that can use environment variables.

Was going to deal with this sed via a common entrypoint, but the script is a much nicer approach. Left the attributes in place for use in the magento1 harness for updating local.yml.


src/_base/helm/app/templates/service/mailhog-relay/service.yaml
```diff
@@ -0,0 +1,11 @@
+{{ if .Values.docker.services.mailhog_relay.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app.service: {{ .Values.resourcePrefix }}mailhog-relay
+  name: {{ .Values.resourcePrefix }}mailhog-relay
```
kierenevans: I guess technically to match the development environment:
```suggestion
  name: mailhog-relay
```

kierenevans:
We have the ability to run entrypoint tasks in console and php-fpm now, so perhaps this should be fixed by running a sed on the php.ini file instead.

tkotosz (approved):
:tada:

PR #447, opened 2020-10-07
Lock the git push for chart publishing

kierenevans:
Should fix this that can happen with parallel builds:
```
 ! [remote rejected] master -> master (cannot lock ref 'refs/heads/master': is at 8fd8d71fb9ec39e94fa92161b9d0213932260d29 but expected fe30974a1f4455b5caddc7c089f7602a738a023b)

error: failed to push some refs to 'git@github.com:example/example.git`
```

### DIFF
--- a/src/_base/application/overlay/Jenkinsfile.twig
+++ b/src/_base/application/overlay/Jenkinsfile.twig
@@ -56,7 +56,9 @@ pipeline {
                 milestone(50)
                 sh 'ws app publish'
 {% if @('pipeline.publish.chart.enabled') %}
-                sh 'ws app publish chart "${GIT_BRANCH}" "{{ @('workspace.name') }} build artifact ${GIT_COMMIT}"'
+                lock(resource: '{{ @('workspace.name') }}-publish-chart', inversePrecedence: true) {
+                    sh 'ws app publish chart "${GIT_BRANCH}" "{{ @('workspace.name') }} build artifact ${GIT_COMMIT}"'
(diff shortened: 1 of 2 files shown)


src/_base/application/overlay/Jenkinsfile.twig
```diff
@@ -56,7 +56,9 @@ pipeline {
                 milestone(50)
                 sh 'ws app publish'
 {% if @('pipeline.publish.chart.enabled') %}
-                sh 'ws app publish chart "${GIT_BRANCH}" "{{ @('workspace.name') }} build artifact ${GIT_COMMIT}"'
+                lock(resource: '{{ @('workspace.name') }}-publish-chart', inversePrecedence: true) {
```
andytson-inviqa: since there are potentially multiple projects released to a cluster repository, would it be better to use the repository itself in the lock name? otherwise it only prevents individual projects

andytson-inviqa: or, if we always keep the same folder structure in jenkins, the folder path

kierenevans: Good point, will update :+1:

kierenevans: Updated

kierenevans: One more file!

andytson-inviqa (approved):

PR #449, opened 2020-10-07
Magento 2 cron needs zombies to exist

kierenevans:
1. Main cron process spawns new standalone processes as children
2. Main cron process exits after doing it's tasks
3. Zombies created out of the forked processes
4. PID 1 (tini) becomes parent of zombies
5. PID 1 (tini) reaps the zombies
6. Standalone processes never complete

### DIFF
--- a/src/magento2/harness.yml
+++ b/src/magento2/harness.yml
@@ -153,7 +153,7 @@ attributes:
         - run magento cache:clean
     cron:
       jobs:
-        - "= (@('app.mode') == 'production' ? '* * * * * /cron-run-with-env.sh /bin/bash -c \"/app/bin/magento cron:run & && wait\"' : '')"
+        - "= (@('app.mode') == 'production' ? '* * * * * /cron-run-with-env.sh \"/app/bin/magento cron:run & wait\"' : '')"
   docker:
     image:
       console: "= 'my127/magento2:' ~ @('php.version') ~ '-fpm-' ~ (@('php.version') >= 7.4 ? 'buster' : 'stretch') ~ '-console'"
--- a/src/magento2/harness.yml
+++ b/src/magento2/harness.yml
@@ -153,7 +153,7 @@ attributes:
         - run magento cache:clean
     cron:
       jobs:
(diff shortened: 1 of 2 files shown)


tkotosz (dismissed):

andytson-inviqa:
if a bash shell for the cron task could run the task then `wait` (waits for all children to finish) then maybe no need

kierenevans:
I think that's what we did in https://github.com/continuouspipe/dockerfiles/pull/180/files with supervisor PID 1, will adjust.

andytson-inviqa:
I think that's just for allowing shell and not just a single command

more like (maybe) `bash -c '... &  && wait'`. I don't know if the cron.php's forks become the bash process's children tho

kierenevans:
Unfortunately they don't: https://github.com/magento/magento2/blob/2.4.0/app/code/Magento/Cron/Observer/ProcessCronQueueObserver.php#L255-L261

kierenevans:
We'll try this where we've been seeing issues :crossed_fingers:

andytson-inviqa:
that's spawned children rather than forks btw

src/magento2/harness.yml
```diff
@@ -153,7 +153,7 @@ attributes:
         - run magento cache:clean
     cron:
       jobs:
-        - "= (@('app.mode') == 'production' ? '* * * * * /cron-run-with-env.sh /app/bin/magento cron:run' : '')"
+        - "= (@('app.mode') == 'production' ? '* * * * * /cron-run-with-env.sh /bin/bash -c \"/app/bin/magento cron:run & && wait\"' : '')"
```
kierenevans:
```
bash: -c: line 0: syntax error near unexpected token `&&'
bash: -c: line 0: `/bin/bash -c /app/bin/magento cron:run & && wait'
```

kierenevans:
```suggestion
        - "= (@('app.mode') == 'production' ? '* * * * * /cron-run-with-env.sh /bin/bash -m -c \\'\"/app/bin/magento cron:run & wait\"\\'' : '')"
```

tkotosz (approved):

PR #452, opened 2020-10-08
Deploy supporting services in wave 4

kierenevans:
Sealed secrets now in wave 1, so we can't deploy anything relying on secrets in wave 0.

PVC's should be fine in wave 0, same for services and ingresses.


tkotosz (approved):
Tested on a project, working fine :+1:

PR #453, opened 2020-10-08
Varnish - Add checksum of configmap to the pods

kierenevans:
To force re-deploy when the configmap changes

### DIFF
--- a/src/_base/helm/app/templates/service/varnish/statefulset.yaml
+++ b/src/_base/helm/app/templates/service/varnish/statefulset.yaml
@@ -15,6 +15,8 @@ spec:
   template:
     metadata:
       creationTimestamp: null
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/service/varnish/configmap.yaml") . | sha256sum }}
       labels:
         app.service: {{ .Values.resourcePrefix }}varnish
     spec:


src/_base/helm/app/templates/service/varnish/statefulset.yaml
```diff
@@ -15,6 +15,8 @@ spec:
   template:
     metadata:
       creationTimestamp: null
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/service/varnish/configmap.yaml") . | sha256sum }}
```
andytson-inviqa, line 19: some have experienced the sha256 is too long for an annotation, but hopefully that is just in a k8s version lower than EKS 1.17

kierenevans: Could do `sha1sum`?

andytson-inviqa: seems to have changed 4 years ago to 256 characters. sha256 is 64 (where limit seemed to trigger as if annotation was 65)

andytson-inviqa: k8s 1.1 was the version that size limit was released in, so I think we're safe

andytson-inviqa (approved):

PR #454, opened 2020-10-08
Add support for MySQL 8

kierenevans:
Ports across overrides by @dcole-inviqa made for akeneo projects and adds some backwards compatibility.

### DIFF
--- a/src/_base/harness/config/functions.yml
+++ b/src/_base/harness/config/functions.yml
@@ -85,7 +85,6 @@ function('filter_local_services', [services]): |
     $filteredService = [];
     foreach ($service as $key => $value) {
       switch ($key) {
-        case 'arguments':
         case 'enabled':
         case 'environment':
         case 'image':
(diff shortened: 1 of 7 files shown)


src/_base/harness/config/functions.yml
```diff
@@ -85,6 +85,7 @@ function('filter_local_services', [services]): |
     $filteredService = [];
     foreach ($service as $key => $value) {
       switch ($key) {
+        case 'arguments':
```
andytson-inviqa: since this key is quite arbitrary can you instead put it in as:
```
pipeline:
  base:
    services:
      mysql:
        arguments: =@('services.mysql.arguments')
```

kierenevans: Maybe `command_arguments`?

kierenevans: Updated

andytson-inviqa: ok we shouldn't need this line with that change

kierenevans: Ah-ha, will remove :+1:

andytson-inviqa (dismissed):

andytson-inviqa (approved):

PR #456, opened 2020-10-08
Deployable tideways daemon

kierenevans:
* Switch to https://github.com/inviqa/docker-tideways
* Deploy tideways daemon per project in kubernetes

We build with the tideways ini settings present in the php-fpm and console docker images. We delete this .ini file on container startup if the TIDEWAYS_ENABLED environment variable is set to "false". We don't set this environment variable on the development environment to ensure `ws feature tideways on` is the only gate locally.
This avoids needing to do templating logic for enabling it on the fly.

This is also good for consistency across environments - to turn this on in pipeline, you'd also need it in the development environment.


src/_base/harness/attributes/common.yml
```diff
@@ -218,9 +218,9 @@ attributes.default:
       config:
         agent_socket: "tcp://blackfire:8707"
     ext-tideways:
-      enable: no
+      enable: "= ('tideways' in @('app.services') ? 'tideways' : '')"
```
andytson-inviqa:
```suggestion
      enable: "= @('services.tideways.enabled')"
```

src/_base/harness/attributes/common.yml
```diff
@@ -218,9 +218,9 @@ attributes.default:
...
       cli:
-        enable: no
+        enable: "= ('tideways' in @('app.services') ? 'tideways' : '')"
```
andytson-inviqa:
```suggestion
        enable: "= @('services.tideways.enabled')"
```

src/_base/harness/attributes/docker-base.yml
```diff
@@ -167,6 +175,12 @@ attributes:
           enabled: false
         mysql:
           enabled: false
+        php-base:
+          environment:
+            TIDEWAYS_ENABLED: "= @('php.ext-tideways.enable') ? 'true' : 'false'"
```
andytson-inviqa: maybe this but in pipeline.base.services.php-base.environment:
```
            TIDEWAYS_ENABLED: "{{ .Values.service.tideways | default .Values.docker.services.tideways.enabled }}"
```
the deployment template templates these, and quotes, which even turns booleans to quoted string

src/_base/harness/attributes/docker-base.yml
```diff
@@ -153,6 +158,9 @@ attributes:
           metricsEnabled: true
           metricsEndpoints:
             - port: php-fpm-metrics
+        tideways:
```
andytson-inviqa, line 171: I don't know if it's a good idea to implicitly enable tideways for all environments just because it's being used locally. Making it an explicit choice doesn't sound too bad

src/_base/harness/attributes/docker-base.yml
```diff
@@ -167,6 +175,12 @@ attributes:
           enabled: false
         mysql:
           enabled: false
+        php-base:
+          environment:
+            TIDEWAYS_ENABLED: "= @('php.ext-tideways.enable') ? 'true' : 'false'"
+        console:
+          environment:
+            TIDEWAYS_ENABLED: "= @('php.ext-tideways.cli.enable') ? 'true' : 'false'"
```
andytson-inviqa: ditto with the other comment but perhaps with the cli.enabled symfony expression stuff still

andytson-inviqa (changes requested):

src/_base/harness/attributes/docker-base.yml
andytson-inviqa: there's a cost factor as well right? /cc @rgpjones (just remember this is a public repo so don't talk about internal Inviqa stuff)

src/_base/harness/attributes/common.yml
kierenevans: Is services.tideways.enabled suitable for pipeline usage?

andytson-inviqa: not exactly, and`app.services` isn't either, but see my other comment about using helm .Values.docker.services.tideways to control that

src/_base/harness/attributes/docker-base.yml
kierenevans: Updated :+1:

src/_base/harness/attributes/docker-base.yml
kierenevans: Updated :+1:

src/_base/harness/attributes/docker-base.yml
kierenevans: Updated :+1:

andytson-inviqa (approved):

PR #464, opened 2020-10-15
Prompt for the community or enterprise edition for Akeneo when creating a project

kierenevans:
Requires https://github.com/my127/workspace/pull/63 to add file encryption support to `ws create`.
It will work without that being present in the user's local `ws` tool. It just won't ask any the questions that have a null value in the harness already.

* Prompt for the edition of Akeneo (community/enterprise)
* Avoid needing to cancel the installation when creating a new project and updating composer.json when wanting to use enterprise edition
* Prompt for the project name in Akeneo's portal when using enterprise edition (can't stop this from happening if you answer "community")
* Prompt for SSH public key and a path to the SSH private key file to download from Akeneo's git repository.

### DIFF
--- a/src/akeneo/README.md
+++ b/src/akeneo/README.md
@@ -28,8 +28,8 @@ git commit -m "Initial commit"
 - A valid subscription with Akeneo to gain access to the enterprise edition codebase.
 - A dedicated user in the Akeneo Portal for the purposes of installing the enterprise edition with composer.
   https://help.akeneo.com/portal/articles/get-akeneo-pim-enterprise-archive.html
-- A public and private SSH keypair generated specifically for the dedicated user in the Akeneo Portal, with the public
-  key associated with the dedicated user.
+- A public and private SSH keypair generated for the purposes of installing the codebase, with the public
+  key associated with the dedicated user in the Akeneo Portal.
 
 ### Installation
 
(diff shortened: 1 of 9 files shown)


andytson-inviqa:
does Akeneo Enterprise allow own-generated ssh keys? If not, do they support https token authentication?

kierenevans:
Looks like it's only available via SSH (talking to port 443). I've added a line to the enterprise part of the akeneo README explaining to generate a new keypair to use for the workspace environment. They are similar to github in that you need to associate a key to the account.

src/akeneo/README.md
```diff
@@ -28,6 +28,8 @@ git commit -m "Initial commit"
 - A valid subscription with Akeneo to gain access to the enterprise edition codebase.
 - A dedicated user in the Akeneo Portal for the purposes of installing the enterprise edition with composer.
   https://help.akeneo.com/portal/articles/get-akeneo-pim-enterprise-archive.html
+- A public and private SSH keypair generated specifically for the dedicated user in the Akeneo Portal, with the public
+  key associated with the dedicated user.
```
andytson-inviqa:
```suggestion
- A public and private SSH keypair generated, with the public key associated with the dedicated user in the Akeneo Portal.
```
Previously it reads to me as if `specifically` means it can't be used for other purposes as well

kierenevans: You can associate multiple ssh keys in the portal, so I don't see the harm in having a key just for pulling down code

andytson-inviqa: the opposite I mean, if another repository needs an ssh key (e.g. just one github repository's deploy key) this key is then not specifically for the akeneo repo

kierenevans: Ah I see, we could:
```suggestion
- A public and private SSH keypair generated for the purposes of installing the codebase, with the public
  key associated with the dedicated user in the Akeneo Portal.
```

andytson-inviqa: ok

andytson-inviqa (approved):

PR #465, opened 2020-10-16
Rewrite local.xml on container startup

kierenevans:
Stops local.xml being compiled into the docker image with secrets only valid for the local environment.


andytson-inviqa:
`@('magento.environment')` as well?

a shame after that it's still one more twig block for admin path, otherwise wouldn't need twig (tho if it wasn't EOL, there'd be a valid reason to do that one as well)

kierenevans:
Ah yes, will do :+1:

Still need twig for memcached/redis-session switching

andytson-inviqa:
still seem to be missing `@('magento.environment')`

kierenevans:
Oops, fixed!

andytson-inviqa (approved):

PR #467, opened 2020-10-16
Fix Magento 1 app init not running migrations

kierenevans:
Seems to be when we introduced `app migrate` for deployments in https://github.com/inviqa/harness-base-php/pull/120

### DIFF
--- a/src/magento1/harness.yml
+++ b/src/magento1/harness.yml
@@ -89,9 +89,7 @@ attributes:
       steps:
         - run n98-magerun.phar config:set 'web/unsecure/base_url' "https://${APP_HOST}/"
         - run n98-magerun.phar config:set 'web/secure/base_url'   "https://${APP_HOST}/"
-        - run n98-magerun.phar sys:setup:incremental -n
-        - run n98-magerun.phar index:reindex:all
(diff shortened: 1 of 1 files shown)


src/magento1/harness.yml
```diff
@@ -89,6 +89,9 @@ attributes:
       steps:
         - run n98-magerun.phar config:set 'web/unsecure/base_url' "https://${APP_HOST}/"
         - run n98-magerun.phar config:set 'web/secure/base_url'   "https://${APP_HOST}/"
+        - run n98-magerun.phar sys:setup:incremental -n
+        - run n98-magerun.phar index:reindex:all
+        - run n98-magerun.phar cache:clean
```
andytson-inviqa: would `- app migrate` be too odd alternative?

andytson-inviqa (dismissed):

src/magento1/harness.yml
kierenevans: I was thinking projects may override the migrate steps to do something they don't want to happen on init, but I doubt that will happen. Will update!

kierenevans: Updated

PR #468, opened 2020-10-16
Add Magento 2 var/export PersistentVolumeClaim

kierenevans:
To be able to generate reports via cron that are downloadable via admin, the var/export folder should be shared between cron and php-fpm as well as being persistent.

### DIFF
--- a/src/magento2/harness.yml
+++ b/src/magento2/harness.yml
@@ -32,6 +32,7 @@ attributes:
       - redis-session
       - elasticsearch
     web_directory: /app/pub
+    export_directory: /app/var/export
     media_directory: = @('app.web_directory') ~ '/media'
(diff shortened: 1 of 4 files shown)


tkotosz (commented):
Is this working even if the folder didn't exist? I remember previously I had issues with that the folder didn't even exist in var initially.

tkotosz (dismissed):

kierenevans:
In anything using the PersistentVolumeClaim, the directory will exist and be mounted into the containers fine.
For local, I'm unsure, so have added it as an explicit build step :+1:

tkotosz (dismissed):
thanks

andytson-inviqa (approved):

PR #470, opened 2020-10-16
Make services.*.enabled able to disable all services (including webapp)

andytson-inviqa:
As part of this, also webapp (php+nginx) is also able to be disabled

This changes most code to use services.*.enabled rather than * in app.services

app.services is not deprecated, so still will continue to function as the primary way to set harness default services, and allowable in projects, however the services.*.enabled way is perferrable now in projects, and has the ability to override harness defaults individually per service


src/magento2/application/overlay/app/etc/env.php.twig
```diff
@@ -45,7 +45,7 @@ return [
         ]
     ],
 {% endif -%}
-{% if 'varnish' in @('app.services') and @('replicas.varnish') > 0 -%}
+{% if @('services.rabbitmq.enabled') and @('replicas.varnish') > 0 -%}
```
kierenevans:
```suggestion
{% if @('services.varnish.enabled') and @('replicas.varnish') > 0 -%}
```

src/_base/helm/app/templates/application/webapp/ingress-istio.yaml
```diff
@@ -1,4 +1,4 @@
-{{ if eq .Values.ingress.type "istio" }}
+{{ if and (.Values.docker.services.nginx.enabled) (eq .Values.ingress.type "istio") }}
```
kierenevans: Needs a check for varnish as well as nginx due to `.Values.ingress.target_service` target. Whilst doubtful you'd have varnish and no nginx, varnish's VCL can be doing anything.

src/_base/helm/app/templates/application/webapp/ingress.yaml
```diff
@@ -1,4 +1,4 @@
-{{ if eq .Values.ingress.type "standard" }}
+{{ if and (.Values.docker.services.nginx.enabled) (eq .Values.ingress.type "standard") }}
```
kierenevans: Needs a check for varnish as well as nginx due to `.Values.ingress.target_service` target. Whilst doubtful you'd have varnish and no nginx, varnish's VCL can be doing anything.

src/_base/helm/app/templates/application/webapp/podmonitor.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.prometheus.podMonitoring -}}
+{{- if and (index .Values.docker.services "php-fpm" "enabled") .Values.prometheus.podMonitoring -}}
```
kierenevans: Need to check for nginx or php-fpm due to lines 13-15

kierenevans (changes requested):

src/_base/helm/app/templates/application/webapp/podmonitor.yaml
andytson-inviqa: nginx and php-fpm rather than or, as nginx just uses php-fpm for it's metrics. If a project just uses nginx then maybe can just overwrite the resource template

kierenevans (approved):

hgajjar (approved):

PR #472, opened 2020-10-20
Customisable static asset language and theme deployments

### LINKED ISSUE #421: [magento2] Allow customising the languages to build static assets for
https://github.com/inviqa/harness-base-php/blob/0.10.x/src/magento2/harness.yml#L76 defaults to just building the en_US locale.

We've been having to customise this on a per-project basis like so to get en_GB built and deployed:
```bash
      passthru bin/magento setup:static-content:deploy en_GB -j 8 --theme custom/default --no-interaction --ansi
      passthru bin/magento setup:static-content:deploy en_GB en_US -j 8 --theme Magento/backend --no-interaction --ansi
```

kierenevans:
Fixes #421

### DIFF
--- a/src/magento2/.ci/sample-static/workspace.yml
+++ b/src/magento2/.ci/sample-static/workspace.yml
@@ -19,6 +19,19 @@ attribute('app.repository'): null
 attribute('aws.id'): null
 attribute('aws.key'): null
 
+attribute('magento.static_content'):
+  backend:
+    languages:
+      - en_GB
+      - en_US
+    themes:
+      - Magento/backend
+  frontend:
+    languages:
+      - en_GB
+    themes:
+      - Magento/luma
+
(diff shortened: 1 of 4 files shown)


src/magento2/.ci/sample-static/workspace.yml
```diff
@@ -19,6 +19,19 @@ attribute('app.repository'): null
 attribute('aws.id'): null
 attribute('aws.key'): null
 
+attribute('magento.static_content'):
+  backend:
+    languages:
+      - en_GB
+      - en_US
```
andytson-inviqa, line 26: I know we can't turn off stores in spryker harness, but with the inherent array merge issues should we solve it for new work?

andytson-inviqa: or @dcole-inviqa could we have help in working out a fix to workspace itself?

andytson-inviqa (changes requested):
need to be able to specify overridden single language. Theme is already a single item array so _currently_ unaffected

src/magento2/.ci/sample-static/workspace.yml
kierenevans: Ah, I just re-read this comment. We define as empty in the harness to avoid this problem. This is for testing if it works in Jenkins with some defined settings. Dynamic one is deliberately empty to test that scenario too (which compiles all languages and themes).

andytson-inviqa (approved):

src/magento2/.ci/sample-static/workspace.yml
kierenevans: Have created https://github.com/my127/workspace/issues/64 to track this, as I know it's been bugging everyone

PR #475, opened 2020-10-23
Add persistence for Spryker jenkins home

andytson-inviqa:
It has job logs in that shouldn't be lost on rescheduling, and spryker app-migrate doesn't add jobs back in on new deployments

### DIFF
--- a/src/spryker/helm/app/templates/service/jenkins/deployment.yaml
+++ b/src/spryker/helm/app/templates/service/jenkins/deployment.yaml
@@ -13,11 +13,17 @@ spec:
   selector:
     matchLabels:
       app.service: {{ $.Values.resourcePrefix }}jenkins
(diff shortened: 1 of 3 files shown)


src/spryker/helm/app/templates/service/jenkins/deployment.yaml
```diff
@@ -13,11 +13,17 @@ spec:
   selector:
     matchLabels:
       app.service: {{ $.Values.resourcePrefix }}jenkins
+{{- if $.Values.persistence.postgres.enabled }}
```
kierenevans:
```suggestion
{{- if $.Values.persistence.jenkins.enabled }}
```

kierenevans (dismissed):

kierenevans (approved):

PR #478, opened 2020-10-29
Fix solr data path

kierenevans:
* Different solr versions store data at different paths.
* Custom config for solr 4.x was not being picked up

### DIFF
--- a/src/_base/_twig/docker-compose.yml/service/solr.yml.twig
+++ b/src/_base/_twig/docker-compose.yml/service/solr.yml.twig
@@ -24,7 +24,7 @@
 {% endif %}
     volumes:
 {% if @('services.solr.config_path') %}
-      - {{ @('services.solr.config_path')}}:/opt/solr/server/solr/configsets/{{ @('services.solr.environment.SOLR_CORE_NAME') }}
+      - {{ @('services.solr.config_path')}}:/opt/solr/server/solr/configsets/{{ @('services.solr.environment.SOLR_CORE_NAME') }}/conf
 {% endif %}
 {% if @('services.solr.major_version') == 4 %}
       - solr_data:/opt/solr/example/solr/{{ @('services.solr.environment.SOLR_CORE_NAME') }}/data/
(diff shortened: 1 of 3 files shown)


src/_base/_twig/docker-compose.yml/service/solr.yml.twig
```diff
@@ -20,13 +20,19 @@
...
+      - {{ @('services.solr.config_path')}}:/opt/solr/server/solr/configsets/{{ @('services.solr.environment.SOLR_CORE_NAME') }}/conf
+{% endif %}
+{% if @('services.solr.major_version') == 4 %}
+      - solr_data:/opt/solr/example/solr
+{% elseif @('services.solr.major_version') < 8 %}
+      - solr_data:/opt/solr/server/solr/mycores
+{% else %}
+      - solr_data:/var/solr
 {% endif %}
-      - solr_data:/var/solr/{{ @('services.solr.environment.SOLR_CORE_NAME') }}/data/
```
andytson-inviqa, line 29: as solr_data is a named volume, this will error as there's no way to specify named volumes without the docker-compose.yml.twig also being aware of it to add a volumes section to define the named volume.

Looks like a bug in the target branch as well

andytson-inviqa, line 29: as solr_data is a named volume, this will error as there's no way to specify named volumes without the docker-compose.yml.twig also being aware of it to add a volumes section to define the named volume.

Looks like a bug in the target branch as well

andytson-inviqa (changes requested):

src/_base/_twig/docker-compose.yml/service/solr.yml.twig
kierenevans: Defined here: https://github.com/inviqa/harness-base-php/blob/0.11.x/src/_base/docker-compose.yml.twig#L40-L42

andytson-inviqa: oh, missed it. Is it needed though vs a `- path`?

kierenevans: Yeah, sadly

andytson-inviqa (approved):

PR #482, opened 2020-11-04
Trust workspace or pipeline host for drupal

kierenevans:
Using APP_HOST environment variable, configure drupal's trusted host settings.

Documentation: https://www.drupal.org/docs/installing-drupal/trusted-host-settings


andytson-inviqa (approved):
Just a little concerned about backwards compatibility and how multi-site hosts would handle it

kierenevans:
Most projects I've seen are customising this file in their overlay directory anyway. Sometimes to only add the trusted host pattern.

PR #490, opened 2020-11-25
Fix magento2 strip config

### LINKED ISSUE #270: [magento2] strip-magento-config removes minification settings
The following should persist in `config.php` after `strip-magento-config` runs during the install process for magento2:
```
    'system' => [
        'default' => [
            'general' => [
                'locale' => [
                    'code' => 'en_GB'
                ]
            ],
            'dev' => [
                'js' => [
                    'minify_files' => '1',
                    'minify_exclude' => '
                    /tiny_mce/
                '
                ],
                'css' => [
                    'minify_files' => '1',
                    'minify_exclude' => '
                    /tiny_mce/
                '
                ],
                'static' => [
                    'sign' => '1'
                ]
            ]
        ]
    ]
```

kierenevans:
Fixes https://github.com/inviqa/harness-base-php/issues/270

### DIFF
--- a/src/magento2/application/skeleton/app/etc/config.php
+++ b/src/magento2/application/skeleton/app/etc/config.php
@@ -504,6 +504,38 @@ return [
             'code' => 'Magento/luma'
         ]
     ],
+    'system' => [
+        'default' => [
+            'general' => [
+                'locale' => [
+                    'code' => 'en_GB'
+                ]
+            ],
+            'dev' => [
+                'js' => [
+                    'merge_files' => '0',
+                    'minify_files' => '0',
+                    'minify_exclude' => [
+                        'tiny_mce' => '/tiny_mce/',
+                        'cardinal_commerce' => '/v1/songbird'
+                    ]
+                ],
+                'css' => [
+                    'minify_files' => '0',
+                    'minify_exclude' => [
+                        'tiny_mce' => '/tiny_mce/'
+                    ],
+                    'use_css_critical_path' => '0'
+                ],
+                'static' => [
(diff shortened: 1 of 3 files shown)


andytson-inviqa:
This seems to break the build atm on the magento2 steps

kierenevans:
The 500 response during behat was down to php-fpm's docker container having the original config.php, and console's docker container having the config.php that was the result of `app:config:dump`.

Magento running via php-fpm detected a difference and refused to serve the application.

Usually project setups are done in dynamic mode where config.php will be the same in all containers via a mountpoint, so this only affects our "from nothing" static tests in Jenkins.

andytson-inviqa:
not project CI or pipeline environments though, which are effectively from nothing

src/magento2/application/skeleton/app/etc/config.php
```diff
@@ -504,6 +504,38 @@
...
+                ],
+                'static' => [
+                    'sign' => '1'
+                ],
+                'template' => [
+                    'minify_html' => '0'
+                ]
+            ]
+        ]
+    ],
```
andytson-inviqa, line 538: isn't this then fragile in that any environment that has different configuration will fail?

kierenevans:
I guess this could happen in project CI environments if config.php is not representative of the extra settings.

andytson-inviqa:
and pipeline environments?

kierenevans:
Only when starting an environment up without a database, triggering the installer.

andytson-inviqa (approved):

PR #499, opened 2020-12-01
Configure file based opcache for cli

### LINKED ISSUE #495: Use a file based opcache for CLI tasks in production mode
CLI processes do not have a shared memory area to share between runs, so each time PHP compiles a file to run the CLI task, the bytecode of the compiled PHP scripts are thrown away.

Configuring a file based cache can help with this, though it can be a bit tricky to manage.

We have found success configuring the following to speed up behat scenarios based on the symfony behat extension:
```yaml
attribute('php.cli.ini'):
  opcache.file_cache: "= (@('app.build') == 'static' ? '/tmp/php-file-cache' : '')"
  opcache.file_cache_only: "= (@('app.build') == 'static' ? '1' : '0')"
  opcache.file_cache_consistency_checks: "= (@('app.build') == 'static' ? '1' : '0')"
```
and making sure the build user can edit that directory in Dockerfile steps: `chown -R build:build /tmp/php-file-cache/`

This could be extended from the console container to any cron or runner containers.

kierenevans:
Fixes #495 

* Allow customising php.ini settings in cron, jenkins-runner and job-queue-consumer
* Set opcache.file_cache_only when in static mode for console, cron, jenkins-runner and job-queue-consumer. Static mode applies whilst in Jenkins and running on kubernetes
* Ship /tmp/php-file-cache as part of the console docker image. We might consider applying this to php-fpm too, falling back to the file cache if the in-memory cache is empty. This might open more cans of worms than we want to deal with however.
* Drupal CLI seems incompatible with file cache (segfaults!)

### DIFF
--- a/src/_base/docker/image/console/Dockerfile.twig
+++ b/src/_base/docker/image/console/Dockerfile.twig
@@ -12,7 +12,9 @@ RUN chown -R build:build /home/build \
  && curl --fail --silent --location --output /sbin/tini https://github.com/krallin/tini/releases/download/v0.19.0/tini \
  && chmod +x /sbin/tini \
  && curl --fail --silent --location --output /usr/local/bin/mhsendmail https://github.com/mailhog/mhsendmail/releases/download/v0.2.0/mhsendmail_linux_amd64 \
- && chmod +x /usr/local/bin/mhsendmail
+ && chmod +x /usr/local/bin/mhsendmail \
+ && mkdir -p /tmp/php-file-cache \
+ && chown -R build:build /tmp/php-file-cache
 {%- set install_extensions=@('php.install_extensions')|merge(@('php.cli.install_extensions'))|filter(v => v is not empty) %}
 {%- if install_extensions %} \
  && cd /root/installer \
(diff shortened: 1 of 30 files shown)


hgajjar (commented):
Just a question, Are you adding file based opcache so that it can be shipped with docker image?

kierenevans:
> Just a question, Are you adding file based opcache so that it can be shipped with docker image?

Wasn't the original intention but a nice side effect that we can keep :+1:

hgajjar:
> > Just a question, Are you adding file based opcache so that it can be shipped with docker image?
> 
> Wasn't the original intention but a nice side effect that we can keep 👍

ok, what was the original intention then? (just curious)

kierenevans:
> ok, what was the original intention then? (just curious)

Having a file based opcache storage allows php to not need to compile files each time a CLI command is called.
We have a project that runs multiple behat suites in succession with the symfony2 extension (which boots the kernel rather than going via a browser). The file cache shaves off a several minutes for the runtime.

andytson-inviqa (approved):

PR #500, opened 2020-12-02
Xdebug 3 opt-in & config support

### LINKED ISSUE #491: Upgrade xdebug version to 3.0
Xdebug 3.0 is released now, it has many improvements including performance. Here is the upgrade documentation link for upgrading from xdebug 2.0 to 3.0

https://xdebug.org/docs/upgrade_guide

Persata:
I've been tinkering with some ideas for allowing configuration based switching between Xdebug v2 and v3 and thought I'd open a PR to get some comments.

This updates the `ext-xdebug` configuration to have a version switch, as well as new config keys for the required ini settings for Xdebug 3 as they have all changed.

There's a new command which will check the major version installed on both the `console` and `php-fpm` containers and switch to the desired version if needed, so you should be able to switch by adding:

```yml
attribute('php.ext-xdebug.version'): 3
```

to your `workspace.override.yml` file and then running:

```shell
ws feature xdebug version-sync
```

I've also added this command as an event after `harness.install` so that projects can have Xdebug 3 set as the default, in which case it would update after the initial install. There might be a tidier way of adding this to the `Dockerfile` instead, maybe a conditional such as `{% if @('php.ext-xdebug.version') == '3' %}`, assuming we're keeping the default pinned at 2.9.8 in the base image.

I've set the default `mode` as just `debug` to allow for step-debugging. There's also the option of doing `develop,debug` which adds nicer output to `var_dump` and better exception output, at the cost of some overhead  - about 25% performance hit to enable both when I tested it.

I figured however that in most cases the framework is handling exception printing, and if you've got the Xdebug extension enabled you might not be using `var_dump` anyway. This is configurable by changing the `php.ext-xdebug.config.v3.mode` too, for example `debug,profile`.

The `start_with_request` config setting is also `yes` in the config as we had `remote_autostart` enabled previously, so keeps the same behaviour.

I've tested this by spinning up new sites with the Symfony and Wordpress harnesses so far.

There are likely better ways to handle the above, but worst case the ini settings from this might come in handy :smile: 

~I see this is failing the tests against the static builds containers, I'll try and fix that.~

### DIFF
--- a/src/_base/harness/attributes/common.yml
+++ b/src/_base/harness/attributes/common.yml
@@ -231,15 +231,22 @@ attributes.default:
         sample_rate: 25
         service: web
     ext-xdebug:
+      version: 2
       enable: no
       cli:
         enable: no
       config:
-        remote_enable: 1
-        remote_autostart: 1
-        remote_port: 9000
-        remote_host: host.docker.internal
-        idekey: workspace
+        v2:
+          remote_enable: 1
+          remote_autostart: 1
+          remote_port: 9000
+          remote_host: host.docker.internal
+          idekey: workspace
+        v3:
+          mode: debug
+          start_with_request: yes
+          client_port: 9003
+          client_host: host.docker.internal
 
   assets:
     remote: ="s3://"~@("aws.bucket")~"/development"
--- a/src/_base/harness/attributes/common.yml
+++ b/src/_base/harness/attributes/common.yml
@@ -231,7 +231,7 @@ attributes.default:
         sample_rate: 25
         service: web
     ext-xdebug:
-      version: 2
+      version: "= (@('php.version') >= 8.0 ? '3' : '2')"
       enable: no
       cli:
         enable: no
--- a/src/_base/harness/attributes/common.yml
+++ b/src/_base/harness/attributes/common.yml
@@ -231,7 +231,7 @@ attributes.default:
         sample_rate: 25
         service: web
     ext-xdebug:
-      version: "= (@('php.version') >= 8.0 ? '3' : '2')"
+      version: "= (version_compare(@('php.version'), '8.0', '>=') ? '3' : '2')"
       enable: no
       cli:
         enable: no
--- a/src/_base/harness/scripts/xdebug_version_sync.sh
+++ b/src/_base/harness/scripts/xdebug_version_sync.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+
+xdebug_version_sync()
+{
+  INSTALLED_VERSION=$(docker-compose exec -T -u root "$1" pecl info xdebug 2>&1 | grep "Release Version" | awk '{print $3}')
+  INSTALLED_MAJOR_VERSION=$(echo "$INSTALLED_VERSION" | cut -d '.' -f1)
+
+  if [ "$INSTALLED_MAJOR_VERSION" == "$XDEBUG_VERSION" ]; then
+    echo "Xdebug in $1 is already on the desired major version $XDEBUG_VERSION (found installed version $INSTALLED_VERSION)"
+    return 0
+  fi
+
+  case $XDEBUG_VERSION in
+    "3")
+      INSTALL_CANDIDATE="xdebug"
+      ;;
+    "2")
+      INSTALL_CANDIDATE="xdebug-2.9.8"
+      ;;
+    *)
+      echo "Invalid Xdebug version provided - please check the 'php.ext-xdebug.version' attribute is '2' or  '3'."
+      exit 1
+  esac
+
+  run docker-compose exec -T -u root "$1" pecl uninstall xdebug
+  run docker-compose exec -T -u root "$1" pecl install "$INSTALL_CANDIDATE"
+}
+
+xdebug_version_sync 'console'
+xdebug_version_sync 'php-fpm'
--- a/src/_base/harness/scripts/xdebug_version_sync.sh
+++ b/src/_base/harness/scripts/xdebug_version_sync.sh
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+PREPARE_CONFIG=0
+RESTART_FPM=0
+
 xdebug_version_sync()
 {
   INSTALLED_VERSION=$(docker-compose exec -T -u root "$1" pecl info xdebug 2>&1 | grep "Release Version" | awk '{print $3}')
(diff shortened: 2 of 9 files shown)


kierenevans:
Great job! I really like this implementation.

Do you imagine a whole project team opting into Xdebug v3 (with ability to downgrade to v2 via workspace.override.yml)?

If yes then I think it would be worth doing what you mentioned and adding a check to upgrade into the console and php-fpm Dockerfile.twig's, so it's "baked in" to the image used locally and in Jenkins. This would be similar to a node version override that's currently in the console Dockerfile.twig.

We can then also make the README.md.twig talk about how to downgrade by checking the conditional for v3 again. I would hope it would be quite rare to downgrade but I guess not all tooling supports v3 just yet if people aren't using PHPStorm.

For PHP 8.0, v2 is not compatible so we must use v3. The installation of v3 is dealt with in https://github.com/my127/docker-php/pull/44/files#diff-eb164a047fb6aadd342ea21c4b6bed3be6733ea0ff3888a89a82e8f77da0bf10R48-R49 but we will something like this for `php.ext-debug.version` in this PR (or a follow up): `= (@('php.version') >= 8.0 ? '3' : '2')`

Persata:
Sorry about the delay! December was pretty busy.

> Do you imagine a whole project team opting into Xdebug v3 (with ability to downgrade to v2 via workspace.override.yml)?

I think so, version 3 is much faster than 2 so it would be good to get projects onto it by default.

I've added a new check to the console & FPM Dockerfiles that will upgrade Xdebug if PHP is 7.x and Xdebug 3 is specified as the candidate version, and removed the `after('harness.install'):` hook. I figured that v3 would be installed by default on any PHP 8 images, so no action needed there. Does that look like what you were thinking?

Also added in your suggested check to default to version 3 if PHP 8 is specified in `common.yml`.

src/_base/harness/attributes/common.yml
```diff
@@ -231,15 +231,22 @@ attributes.default:
         sample_rate: 25
         service: web
     ext-xdebug:
+      version: "= (@('php.version') >= 8.0 ? '3' : '2')"
```
kierenevans:
```suggestion
      version: "= (version_compare(@('php.version'), '8.0', '>=') ? '3' : '2')"
```
Based on similar feedback from Andy in https://github.com/inviqa/harness-base-php/pull/506

src/_base/harness/scripts/xdebug_version_sync.sh
```diff
@@ -0,0 +1,48 @@
...
+
+xdebug_version_sync 'console'
+xdebug_version_sync 'php-fpm'
+
+if [ "$PREPARE_CONFIG" == 1 ]; then
+  run ws install --step=prepare
+fi
+
+if [ "$RESTART_FPM" == 1 ]; then
+  run ws service php-fpm restart
```
kierenevans, line 41: [`ws service php-fpm restart`](https://github.com/inviqa/harness-base-php/blob/0.12.x/src/_base/harness/config/commands.yml#L163-L171) also ensures the rendered configuration is copied into the console and php-fpm containers, so would need to run in the case of an upgrade of Xdebug in the console container to get it using the new Xdebug v3 configuration.

src/_base/harness/scripts/xdebug_version_sync.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+
+PREPARE_CONFIG=0
+RESTART_FPM=0
+
+xdebug_version_sync()
+{
+  INSTALLED_VERSION=$(docker-compose exec -T -u root "$1" pecl info xdebug 2>&1 | grep "Release Version" | awk '{print $3}')
```
kierenevans:
```suggestion
  local INSTALLED_VERSION
  INSTALLED_VERSION="$(docker-compose exec -T -u root "$1" pecl info xdebug 2>&1 | grep "Release Version" | awk '{print $3}')"
```

src/_base/harness/scripts/xdebug_version_sync.sh
```diff
@@ -0,0 +1,48 @@
...
+  INSTALLED_MAJOR_VERSION=$(echo "$INSTALLED_VERSION" | cut -d '.' -f1)
```
kierenevans:
```suggestion
  local INSTALLED_MAJOR_VERSION
  INSTALLED_MAJOR_VERSION="$(echo "$INSTALLED_VERSION" | cut -d '.' -f1)"
```

src/_base/harness/scripts/xdebug_version_sync.sh
```diff
@@ -0,0 +1,48 @@
...
+
+  if [ "$INSTALLED_MAJOR_VERSION" == "$XDEBUG_VERSION" ]; then
```
kierenevans:
```suggestion
  if [[ "$INSTALLED_MAJOR_VERSION" == "$XDEBUG_VERSION" ]]; then
```
as `==` is only available in double brackets

src/_base/harness/scripts/xdebug_version_sync.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+
+PREPARE_CONFIG=0
```
kierenevans:
```suggestion
set -e -o pipefail

PREPARE_CONFIG=0
```
To stop execution if any command fails.

kierenevans (commented):
Just a few script related things 👍

src/_base/harness/scripts/xdebug_version_sync.sh
Persata: Ah good catch! I've combined the two variables into just `UPDATE_CONTAINERS` since we would need to run the FPM restart in both cases, and I see the restart command already does `ws install --step=prepare` as part of it.

src/_base/harness/scripts/xdebug_version_sync.sh
Persata: Have added these, and also used `local` on the `INSTALL_CANDIDATE` variable - is that OK too?

src/_base/harness/scripts/xdebug_version_sync.sh
Persata: I've added the double brackets around the check for restarting the container too since I was using `==` there also.

src/_base/harness/scripts/xdebug_version_sync.sh
kierenevans: Perfect :)
Stops the global scope being affected

kierenevans (dismissed):
Awesome work!

kierenevans (approved):

PR #511, opened 2020-12-18
Override zed session cookie domain from zed.de.suite.local

kierenevans:
zed.de.suite.local seems to be the default though I can't see it mentioned in code anywhere.

To be able to log in to zed, the CSRF token needs to match a session value. The session cookie was trying to be set for zed.de.suite.local, so was not then forwarded in the POST request to Zed's login screen.

### DIFF
--- a/src/spryker/application/skeleton/config/Shared/config_local.php
+++ b/src/spryker/application/skeleton/config/Shared/config_local.php
@@ -96,6 +96,7 @@ $config[ApplicationConstants::BASE_URL_SSL_STATIC_MEDIA] = $config[ApplicationCo
 
 // ---------- Session
 $config[SessionConstants::ZED_SESSION_COOKIE_NAME] = $config[ApplicationConstants::HOST_ZED];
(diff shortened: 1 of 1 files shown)


zsotyooo (approved):

hgajjar (approved):
Seems like it was using default value before (blank), but is set to zed.de.suite.local in latest demoshop release https://github.com/spryker-shop/b2c-demo-shop/blob/202009.0/config/Shared/config_default-development.php#L155-L156

We should check if all config parameters using $yvesHost, $zedHost and $glueHost in that config_default-development.php are overwritten in config_local.php, if not then add them too.

kierenevans:
Great idea!

I've done:
```php
var_dump($config);die();
```
at the end of the project's config_local.php and the only reference to .local now is:
```
    ["TESTIFY_CONSTANTS:GLUE_APPLICATION_DOMAIN"]=>
    string(26) "http://glue.de.suite.local"
```

PR #513, opened 2021-01-05
Enable Spryker Jenkins CSRF support

andytson-inviqa:
The Jenkins newer versions require CSRF if not using an api token


kierenevans (dismissed):

andytson-inviqa:
still having failures with this on a project. May be to either disable CSRF or complex (but more secure) api token

PR #514, opened 2021-01-05
Make Jenkins env vars configurable and disable CSRF

andytson-inviqa:
to work around Jenkins upgrade issue


hgajjar (approved):
Looks good, not sure why build is failing, got 522 while trying to check

PR #517, opened 2021-01-06
#67 Detect if containers exist during `ws enable`

### LINKED ISSUE #67: Detect if containers have been deleted without deleting .flag-built
An example after having run `docker system prune -a`:
```
ws install
[/Users/kevans/Repositories/ws/magento2-sample/.my127ws]:
■ > docker-compose -p magento2-sample up -d
Creating network "magento2-sample_private" with the default driver
Building console
Step 1/7 : FROM my127/magento2:7.1-fpm-stretch-console
7.1-fpm-stretch-console: Pulling from my127/magento2
Digest: sha256:6c7928a24cf82a7b09d6b52f9d57d33771131e3378fcead0c5ed1a9e639a7905
Status: Downloaded newer image for my127/magento2:7.1-fpm-stretch-console
 ---> 50dfa4c391dc
Step 2/7 : COPY .my127ws/docker/image/console/root /
 ---> e01eff757498
Step 3/7 : RUN chown -R build:build /home/build
 ---> Running in 3345df6ce70c
```

This has skipped the assets download and the application build due to .my127ws/.flag-built existing from a previous install, but the docker images and containers no longer exist locally.

kierenevans:
If no containers are available for the project but .my127ws/.flag-built still exists, count it as needing to build the project as containers would be brought up without any database volumes.

Fixes #67


src/_base/harness/scripts/enable.sh.twig
```diff
@@ -2,10 +2,23 @@
...
+            IS_BUILT=no
+        fi
+    fi
 
-        passthru docker-compose down
+    passthru ws networks external
+    if [ "$IS_BUILT" = "no" ]; then
+        if [ "$HAS_FLAG" = "no" ]; then
+          passthru docker-compose down
+        fi
```
hgajjar, line 21: So if the setup process is halted by something, it would _retry_ the setup from where it stopped instead of removing everything and starting from scratch?

kierenevans: If the setup process stops, the flag is never created, so would still bring containers down to try again.

hgajjar: ah right, I was thinking about an existing env, where I retry the setup and it stops, but then the retry step would/should have removed the flag, all good 👍

hgajjar (dismissed):

PR #518, opened 2021-01-07
Fix redis-session not enabled on kubernetes

kierenevans:
Standardise on `redis-session` to match `php-fpm`


src/_base/harness/attributes/docker-base.yml
```diff
@@ -190,7 +195,7 @@ attributes:
             RABBITMQ_HOST: '{{ if .Values.service.rabbitmq | default .Values.docker.services.rabbitmq.enabled }}{{ .Values.resourcePrefix  }}rabbitmq{{ end }}'
             RABBITMQ_EXTERNAL_HOST: = @('pipeline.preview.rabbitmq.external_host')
             REDIS_HOST: '{{ if .Values.service.redis | default .Values.docker.services.redis.enabled }}{{ .Values.resourcePrefix }}redis{{ end }}'
-            REDIS_SESSION_HOST: '{{ if .Values.service.redis_session | default .Values.docker.services.redis_session.enabled }}{{ .Values.resourcePrefix }}redis-session{{ end }}'
+            REDIS_SESSION_HOST: '{{ if .Values.service.redis_session | default (index .Values.docker.services "redis-session" "enabled") }}{{ .Values.resourcePrefix }}redis-session{{ end }}'
```
hgajjar: Is `index` required because one of the keys contains `-` now?

src/_base/harness/attributes/common.yml
```diff
@@ -339,6 +341,10 @@ attributes.default:
     host: redis
     port: 6379
 
+  redis-session:
+    host: = @('redis_session.host')
+    port: = @('redis_session.port')
+
   redis_session:
     host: redis-session
     port: 6379
```
hgajjar: why keeping the old reference of redis_session everywhere? for backward compatibility?

src/_base/harness/attributes/docker-base.yml
kierenevans: Yep!

src/_base/harness/attributes/common.yml
kierenevans: Yes, in case someone has customised redis_session variants since 0.10.0
Doubtful but... maybe :D

hgajjar: It's ok, but it just seems confusing. Should we rather have this in major release, where BC breaks are allowed? so we can then remove the old references.

kierenevans: We need a fix for redis-session not being deployed to kubernetes, as it breaks magento 2.
Perhaps we call this 1.0.0?

hgajjar: sounds good 👍

hgajjar (dismissed):
PR looks good now, but drupal build is failing, not sure if it's really the changes of this PR.

andytson-inviqa (dismissed):

kierenevans:
Adjusted for the deprecation removals in #519

andytson-inviqa (approved):

PR #519, opened 2021-01-08
Remove deprecations

kierenevans:
Helm and docker configurations have changed over time and we've maintained backwards compatibility.
As we prepare for v1.0.0, we should remove the backwards compatibility.


src/_base/harness/attributes/docker-base.yml
```diff
@@ -55,6 +55,8 @@ attributes:
         DB_ADMIN_PASS: = @('database.root_pass')
       resources:
         memory: "2048Mi"
+        init_memory: "1024Mi"
+        migrate_memory: "1024Mi"
```
andytson-inviqa, line 59: I've done it another way on a project, but ok with this for now

andytson-inviqa (dismissed):
some more work to do for a 1.0.0 milestone but can be merged into a 1.0.0 branch

src/spryker/_twig/docker-compose.yml/environment.yml.twig
```diff
@@ -1,12 +0,0 @@
-{% for store, vhost in @('rabbitmq.vhosts') %}
```
kierenevans, line 1: We need to keep this file and it's includes or provide another way to have these environment variables present.

kierenevans: Now resolved :balloon:

PR #521, opened 2021-01-12
DEVOPS-1181 lighthouse integration

g-foster2020:


src/_base/docker/image/lighthouse/root/app/run.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+reportDir="/home/headless/results"
+
+mkdir -p "${reportDir}"
+
+lighthouse --no-enable-error-reporting \
+           --chrome-flags="--headless --no-sandbox=true" \
+           --output-path=/home/headless/results/lighthouse-results.html \
+           ${TARGET_URL}
```
kierenevans: shellcheck is complaining that TARGET_URL needs double quoting.

I'd also add `set -e -o pipefail` to after the shebang line to exit on any line erroring.

src/_base/harness/attributes/common.yml
```diff
@@ -157,6 +157,10 @@ attributes.default:
     default_branch: = @('git.main_branch')
     # potentially deprecated, unless it is repurposed
     main_branch: develop
+  
+  lighthouse:
+    target:
+      url: ~
```
kierenevans: Should the default be `= @('hostname')` to pick up on the main URL for the project?

src/_base/docker/image/lighthouse/root/app/run.sh
g-foster2020: Improved and re-tested.

src/_base/harness/attributes/common.yml
g-foster2020: Not sure. What do you think @dcole-inviqa ?

coledave: seems like a sensible default :)

kierenevans (dismissed):

kierenevans (approved):

PR #524, opened 2021-01-14
Fix spryker after deprecations removed

kierenevans:
Dynamically define attributes for per-store environment variables, rather than use a twig include. This allows us to have all environment variables written out to both docker and helm templates.

Spryker's helm values template is now the same as _base, after removing console.enabled (removed in https://github.com/inviqa/harness-base-php/pull/256), fixing resourcePrefix to use the attribute rather than workspace.name and updating the persistence attribute usage to be `pipeline.base.persistence`.

TODO:
- [x] Preview values.yaml updates


hgajjar (approved):
Looks pretty neat 👍

PR #525, opened 2021-01-14
Use the runtime environment APPLICATION_ENV for spryker install

andytson-inviqa:
If it was in installer configuration, it couldn't be overriden.

Build-time mode will use the same mode for installer as for application (local 'development', pipeline 'production')

### DIFF
--- a/src/spryker/application/skeleton/config/install/docker.yml.twig
+++ b/src/spryker/application/skeleton/config/install/docker.yml.twig
@@ -2,9 +2,6 @@
 # We have removed US store from database related steps as original file expects a separate database for US store
 # which is not the case in this setup
 
-env:
-    APPLICATION_ENV: {{ @('spryker.mode') }}
-
 stores:
     - DE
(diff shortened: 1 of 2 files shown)


kierenevans (dismissed):

kierenevans:
Maybe it should default to production in the file, which is overridden at runtime via docker-compose environment variables to be development if needed?

It's what I had to do with https://github.com/inviqa/spryker-sample/blob/main/config/install/docker.yml#L5-L6

andytson-inviqa:
> Maybe it should default to production in the file, which is overridden at runtime via docker-compose environment variables to be development if needed?

it seemed it being in the file ignores the env var hence the change

hgajjar (approved):
@andytson-inviqa you are right, the env var in installation recipe overwrites existing env vars.
@kierenevans not sure how it worked in spryker-sample repo, may be it just uses production mode always and we haven't noticed?

PR #527, opened 2021-01-14
Adjust spryker to use hostname_aliases for nginx

kierenevans:
* Set hostname_aliases to be an array of the external hostnames for glue, yves and zed
* Document the glue URLs in the README
* Use _base's nginx template now that the hostname_aliases is set up
* Fix spacing in the istio ingress
* Use a `with` in the nginx ingress for _base and spryker
* Update spryker to use istio.gateways rather than a hardcoded one.
* Fix spryker ingresses only containing one host "localhost" due to `.Values.services.nginx.environment.FPM_HOST`
* Restrict environment variables processed in spryker ingress to ones containing `_HOST_` in their name


src/spryker/harness.yml
```diff
@@ -86,6 +86,11 @@ attributes:
   elasticsearch:
     image: elasticsearch
     tag: 6.8.12
+  hostname_aliases: "= replace(flatten([
+      @('glue.external_hosts'),
+      @('yves.external_hosts'),
+      @('zed.external_hosts')
+    ]), @('hostname'), @('namespace'))"
```
hgajjar, line 93: what is the value of namespace here? before we weren't doing any replacement in nginx.yml.twig

src/_base/helm/app/templates/application/webapp/ingress.yaml
```diff
@@ -1,27 +1,29 @@
...
     {{ $key }}: {{ $value | quote }}
   {{- end }}
 {{- end }}
   creationTimestamp: null
   labels:
     {{- include "chart.labels" $ | nindent 4 }}
     app.kubernetes.io/component: webapp
-    app.service: {{ .Values.resourcePrefix }}webapp
-  name: {{ .Values.resourcePrefix }}webapp
+    app.service: {{ $.Values.resourcePrefix }}webapp
```
hgajjar, line 14: what does `$` do different here?

kierenevans: We've added a with on line two, which scopes `.` to be `.Values.services.nginx.*`

The `$` refers to the root context (like we had before `with`)

src/spryker/harness.yml
kierenevans: https://github.com/my127/workspace/blob/0.1.x/docs/types/workspace.md says a DNS-friendly version of `@('workspace.name')`, which is from the workspace.yml line 1. It's used for the docker-compose project name.

For example `ci-spryker-sample-dynamic`.

`@('hostname')` is already `@('workspace.name') ~ '.' ~ @('domain')`, so `ci-spryker-sample-dynamic.my127.site`

This is removing the domain part - `.my127.site` as it is re-added when looping over these aliases in templates.

This might be the same result and be clearer:
```suggestion
    ]), '.' ~ @('domain'), '')"
```

kierenevans: Though, it makes it less obvious when comparing with:
```
  yves:
    external_hosts:
      DE: = 'yves-de-' ~ @('hostname')
```

hgajjar: It's ok, I was just not sure why we need that now 👍

src/_base/helm/app/templates/application/webapp/ingress.yaml
hgajjar: ah nice, pardon my poor twig knowledge 😄

hgajjar (commented):
Looks good to me, just not sure about the changes in `src/spryker/helm/app/templates/application/webapp/ingress-istio.yaml`

src/_base/helm/app/templates/application/webapp/ingress.yaml
kierenevans: This one is just helm/gotpl :)

src/spryker/harness.yml
andytson-inviqa: I like the suggestion more I think.

kierenevans: Applied it now after testing earlier :)

src/_base/helm/app/templates/application/webapp/ingress.yaml
```diff
@@ -1,27 +1,29 @@
 {{ if and (index .Values.services .Values.ingress.target_service "enabled") (eq .Values.ingress.type "standard") }}
+{{- with .Values.services.nginx -}}
 apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
-{{- if .Values.services.nginx.ingress.annotations }}
+{{- if .ingress.annotations }}
   annotations:
-  {{- range $key, $value := .Values.services.nginx.ingress.annotations }}
```
andytson-inviqa, line 7: what do you think to moving .Values.services.nginx.ingress.annotations to .Values.ingress.annotations for the 1.0.0 release?

kierenevans: I guess there might be annotations for the istio ingress too, though that can be switched at the values.yaml.twig level

andytson-inviqa: istio virtualservice only has labels atm, but I'd apply the annotation capability to both of them

kierenevans: How does https://github.com/inviqa/harness-base-php/pull/538 look?

kierenevans: It's now part of this PR :)

andytson-inviqa (approved):

PR #528, opened 2021-01-19
Fix lib folder not being readable

kierenevans:
Some combination of Docker Desktop for macOS, Mutagen and adding .my127ws/docker/image/*/root into docker image builds is making it so that /lib/ and other directories are unable to be readable.

This most often manifests itself as a DNS error during `apt update`:
```
W: Failed to fetch http://security.debian.org/debian-security/dists/stretch/updates/InRelease  Temporary failure resolving 'security.debian.org'
```

The theory is currently that .my127ws/docker/image/console/root/lib/ (as well as equivalents for cron and php-fpm) has lost it's read and execute permissions for other users.

Whilst the cause of this permissions issue is unknown, we can reset permissions just before we build to try to avoid the symptom.

harness-akeneo's enable.sh left alone as it's removed in https://github.com/inviqa/harness-base-php/pull/517


andytson-inviqa:
it makes the project folder lose go-rwx on the host, which gets COPYed in the Dockerfile with the same permissions. The lib folder is just one visible effect of this issue

andytson-inviqa (dismissed):
(jedi force) this is not the cause you're looking for

andytson-inviqa:
a more generic catch could help just by warning the user their project directory has insufficient permissions, which is useful beyond just this current workaround

andytson-inviqa (dismissed):

andytson-inviqa (approved):

PR #530, opened 2021-01-21
Remove docker-sync support

kierenevans:
Use mutagen exclusively.

Migration path for anyone using docker-sync currently is:
```bash
docker-sync stop
docker rm "<workspace-project-name>-sync"
ws switch mutagen
```


hgajjar (approved):
Wow, so many traces of it everywhere 👍

kierenevans:
Lots left in the .gitignore and mutagen ignore sections, but I think that's fine :)

PR #531, opened 2021-01-21
Update magento2 VCL to 2.4.1

kierenevans:
From https://github.com/magento/magento2/blob/2.4.1/app/code/Magento/PageCache/etc/varnish6.vcl


jameshalsall (approved):
nice 👍

andytson-inviqa (approved):

PR #533, opened 2021-01-21
Rename AWS Credentials

### LINKED ISSUE #100: Rename aws.id and aws.key and env vars to match aws's standard env var naming
It's not clear from these names what they are, there is no "id", and there are two components of a "key", an ID and secret.

It would be better to use aws' naming conventions e.g.

aws.access_key.id / AWS_ACCESS_KEY_ID
aws.access_key.secret / AWS_SECRET_ACCESS_KEY

Now that workspace is in use though may hamper the rollout unless backwards compatibility is made in the attributes.

kierenevans:
Fixes #100 

Need to use AWS_ID and AWS_KEY environment variables for https://github.com/my127/workspace/blob/0.1.x/home/bin/ws.aws#L7-L17 until that's changed.

### DIFF
--- a/src/_base/harness/config/commands.yml
+++ b/src/_base/harness/config/commands.yml
@@ -96,16 +96,20 @@ command('db console'):
 
 command('assets download'):
   env:
-    AWS_ID:  =@('aws.id')
-    AWS_KEY: =@('aws.key')
+    AWS_ACCESS_KEY_ID: = @('aws.access_key_id')
+    AWS_ID: = @('aws.access_key_id')
+    AWS_KEY: = @('aws.secret_access_key')
+    AWS_SECRET_ACCESS_KEY: = @('aws.secret_access_key')
   exec: |
     #!bash(workspace:/)|@
     passthru ws.aws s3 sync '@('assets.remote')' '@('assets.local')'
 
 command('assets upload'):
   env:
-    AWS_ID:  =@('aws.id')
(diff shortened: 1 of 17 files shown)


src/_base/harness/config/commands.yml
```diff
@@ -96,16 +96,20 @@ command('db console'):
...
   env:
-    AWS_ID:  =@('aws.id')
-    AWS_KEY: =@('aws.key')
+    AWS_ACCESS_KEY_ID: = @('aws.access_key_id')
+    AWS_ID: = @('aws.access_key_id')
+    AWS_KEY: = @('aws.secret_access_key')
+    AWS_SECRET_ACCESS_KEY: = @('aws.secret_access_key')
   exec: |
     #!bash(workspace:/)|@
     passthru ws.aws s3 sync '@('assets.local')' '@('assets.remote')'
```
andytson-inviqa, line 115: A possible interesting sideffect of this is that prior to this, if aws was installed on host, (since ws.aws uses host aws if installed) it would use the host's ~/.aws/config (etc), and presumably now it will use the project's key

kierenevans: I'm actually not sure how it was using the host's config based on https://github.com/my127/workspace/blob/0.1.x/home/bin/ws.aws#L10

andytson-inviqa (approved):

PR #515, opened 2021-01-06
Clean up more tagged images

kierenevans:
When in dynamic mode on people's machines, we always leave behind images tagged `PROJECTNAME-php-fpm:dev` and `PROJECTNAME-console:dev`

In static mode on CI workers, when the docker repository is configured as `example.com/organisation/repository:prefix-` to share the repository between different services, we leave behind images tagged `example.com/organisation/repository:prefix-commithash-console` due to trying to clean up
```
docker images --filter=reference='example.com/organisation/repository:prefix-:aaf8dfc4af641adadadeefefeff343ae-*'
```
which returns no results.

PR #534, opened 2021-01-21
Add a command hook to cleanup built images

andytson-inviqa:
Replacing potential for discrepancy of deleting images based on matching hardcoding

### DIFF
--- a/src/_base/harness/config/cleanup.yml
+++ b/src/_base/harness/config/cleanup.yml
@@ -14,7 +14,9 @@ function('built_images', [services]): |
 command('cleanup built-images'):
   env:
     IMAGES: = built_images(docker_service_images())
+    BUILD_LABEL: = @('workspace.name') ~ ':' ~ @('app.version')
   exec: |
     #!bash
     IMAGES=($IMAGES)
-    docker image rm --force -- "${IMAGES[@]}"
+    run docker image rm --force -- "${IMAGES[@]}"
+    run "docker images --filter=label=build=$(printf '%q' "$BUILD_LABEL") -q | xargs --no-run-if-empty docker image rm --force"
--- a/src/_base/harness/config/cleanup.yml
+++ b/src/_base/harness/config/cleanup.yml
@@ -1,0 +1,19 @@
+function('built_images', [services]): |
+  #!php
+  $builtImages = [];
+
+  foreach ($services as $service) {
+    if ($service['image'] && count($service['upstream']) > 0) {
+      $builtImages[] = $service['image'];
+    }
+  }
+
+  # workspace commands don't allow non-string types
+  = join(' ', $builtImages);
+
+command('cleanup built-images'):
+  env:
+    IMAGES: = built_images(docker_service_images())
+  exec: |
+    #!bash
+    docker image rm --force $IMAGES
--- a/src/_base/harness/config/cleanup.yml
+++ b/src/_base/harness/config/cleanup.yml
@@ -16,4 +16,5 @@ command('cleanup built-images'):
     IMAGES: = built_images(docker_service_images())
   exec: |
     #!bash
-    docker image rm --force $IMAGES
+    IMAGES=($IMAGES)
+    docker image rm --force -- "${IMAGES[@]}"
--- a/src/_base/harness/config/cleanup.yml
+++ b/src/_base/harness/config/cleanup.yml
@@ -18,5 +18,7 @@ command('cleanup built-images'):
   exec: |
     #!bash
     IMAGES=($IMAGES)
-    run docker image rm --force -- "${IMAGES[@]}"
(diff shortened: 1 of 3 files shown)


kierenevans:
Okay to include https://github.com/inviqa/harness-base-php/pull/515's additions?

andytson-inviqa:
can you tell me why they wouldn't be covered by this? (maybe in slack as it'll be a bit of discussion)

kierenevans:
Though it might do already actually!

Sorry for the build failure, trying to replace jenkins workers and had to cancel it

src/_base/harness/config/cleanup.yml
```diff
@@ -0,0 +1,19 @@
...
+
+  # workspace commands don't allow non-string types
+  = join(' ', $builtImages);
+
+command('cleanup built-images'):
+  env:
+    IMAGES: = built_images(docker_service_images())
+  exec: |
+    #!php
+    docker rm $IMAGES
```
kierenevans:
```suggestion
    docker image rm $IMAGES
```

src/_base/harness/config/cleanup.yml
```diff
@@ -0,0 +1,19 @@
...
+  }
+
+  # workspace commands don't allow non-string types
+  = join(' ', $builtImages);
+
+command('cleanup built-images'):
+  env:
+    IMAGES: = built_images(docker_service_images())
+  exec: |
+    #!php
```
kierenevans:
```suggestion
    #!bash
```

kierenevans:
Testing this locally with `./build && ./test php static`, I'm left with a 922MB image:
```bash
docker image ls | head -n30
REPOSITORY                            TAG                        IMAGE ID       CREATED         SIZE
<none>                                <none>                     49581cfebaeb   2 minutes ago   922MB
```
Seems to be due to the multi-stage build for php-fpm from console:
```bash
■ > docker-compose build php-fpm
Building php-fpm
Step 1/13 : FROM ci-php-sample-static:b1908f9e057628fc039b085c6ee94d9b659a5dfb-console as console
 ---> 91b3f0326547
Step 2/13 : RUN if [ -d /app/tools/assets/ ]; then rm -rf /app/tools/assets/; fi
 ---> Running in aad57c2336b5
Removing intermediate container aad57c2336b5
 ---> 442cfdabf896
Step 3/13 : RUN if [ -d /app/node_modules/ ]; then rm -rf /app/node_modules/; fi
 ---> Running in 3cd1e964ed86
Removing intermediate container 3cd1e964ed86
 ---> 49581cfebaeb

Step 4/13 : FROM my127/php:8.0-fpm-buster
 ---> af838a7410dc
Step 5/13 : WORKDIR /app
 ---> Running in 18b4e8e15196
Removing intermediate container 18b4e8e15196
 ---> e07650a89137
 ```

**Comment by andytson-inviqa** [comment]
that's why I'm thinking `docker image prune` in CI, either here or a regular cron. Can also alternatively label the image to delete it by filter, but would need to remember that practice for any other multi-stage build (which may not even be FROM a project's image)

File: src/_base/harness/config/cleanup.yml [hunk 3]
```diff
@@ -0,0 +1,20 @@
...
+  # workspace commands don't allow non-string types
+  = join(' ', $builtImages);
+
+command('cleanup built-images'):
+  env:
+    IMAGES: = built_images(docker_service_images())
+  exec: |
+    #!bash
+    IMAGES=($IMAGES)
+    docker image rm --force -- "${IMAGES[@]}"
```
kierenevans: This seems to catch the stragglers without needing to prune, though it deletes nginx and php-fpm early and then complains they don't exist later:
```suggestion
    for image in ${IMAGES[@]}; do
      docker images --filter=since="${image}" -q | xargs --no-run-if-empty docker image rm --force
    done
    docker image rm --force -- "${IMAGES[@]}"
```

This is currently running in dynamic mode too so need to drop `--no-run-if-empty` if on macOS at the risk of an error occurring if no entries:

```suggestion
    for image in ${IMAGES[@]}; do
      docker images --filter=since="${image}" -q | xargs docker image rm --force
    done
    docker image rm --force -- "${IMAGES[@]}"
```

andytson-inviqa: I think a prune is fine, it only deletes dangling images (of which would only be multi-stage builds and previous build failures not cleaned up) anyway

andytson-inviqa: only the LABEL way tho would be suitable for concurrent executors, both filter=since and prune would have issues

andytson-inviqa: I've added the LABEL way as is good for local too, but I think if there's good enough reason for anything more then for CI it be prune as is simple with the same consequences as complex filter=since, but even better because it cleans up all prior dangling images as well

kierenevans (approved):

PR #538, opened 2021-01-26
Move ingress annotations

kierenevans:
* Enable istio by setting `attribute('pipeline.base.istio.enabled'): true`.
* Switch between nginx and istio ingress annotations in values.yaml.twig to provide `.Values.ingress.annotations`
* 
Merges into #527

### DIFF
--- a/src/_base/harness/attributes/docker-base.yml
+++ b/src/_base/harness/attributes/docker-base.yml
@@ -221,7 +221,9 @@ attributes:
             TIDEWAYS_HOSTNAME: = @('pipeline.base.hostname')
       ingress:
         annotations: {}
-        type: standard # standard or istio
+        target_service: "= @('services.varnish.enabled') ? 'varnish' : 'webapp'"
+        # standard or istio
+        type: standard
       istio:
         gateways:
           - "istio-system/{{ .Release.Namespace }}-gateway"
--- a/src/_base/harness/attributes/docker-base.yml
+++ b/src/_base/harness/attributes/docker-base.yml
@@ -221,8 +221,8 @@ attributes:
             TIDEWAYS_HOSTNAME: = @('pipeline.base.hostname')
       ingress:
         annotations: {}
+        type: standard # standard or istio
       istio:
-        enabled: false
         gateways:
           - "istio-system/{{ .Release.Namespace }}-gateway"
         additionalGateways: []
--- a/src/_base/harness/attributes/docker-base.yml
+++ b/src/_base/harness/attributes/docker-base.yml
@@ -222,9 +222,12 @@ attributes:
           environment:
             TIDEWAYS_HOSTNAME: = @('pipeline.base.hostname')
       istio:
+        enabled: false
         gateways:
           - "istio-system/{{ .Release.Namespace }}-gateway"
         additionalGateways: []
+        ingress:
+          annotations: {}
     production:
       # assumption is that in a production style environment these would be
       # managed services outside of the applications control
--- a/src/_base/harness/attributes/docker-base.yml
+++ b/src/_base/harness/attributes/docker-base.yml
@@ -205,8 +205,6 @@ attributes:
         nginx:
           environment:
             FPM_HOST: localhost
-          ingress:
-            annotations: {}
           metricsEnabled: false
           metricsEndpoints:
             - port: http
@@ -221,13 +219,13 @@ attributes:
         tideways:
           environment:
             TIDEWAYS_HOSTNAME: = @('pipeline.base.hostname')
+      ingress:
+        annotations: {}
       istio:
         enabled: false
         gateways:
           - "istio-system/{{ .Release.Namespace }}-gateway"
         additionalGateways: []
-        ingress:
-          annotations: {}
     production:
       # assumption is that in a production style environment these would be
       # managed services outside of the applications control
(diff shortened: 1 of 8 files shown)


src/_base/helm/app/templates/application/webapp/ingress-istio.yaml
```diff
@@ -3,6 +3,12 @@ apiVersion: networking.istio.io/v1alpha3
 kind: VirtualService
 metadata:
   name: {{ .Release.Namespace }}-{{ .Values.resourcePrefix }}webapp-virtualservice
+{{- if .Values.ingress.annotations }}
+  annotations:
+  {{- range $key, $value := .Values.ingress.annotations }}
+    {{ $key }}: {{ $value | quote }}
+  {{- end }}
+{{- end }}
```
andytson-inviqa:
```suggestion
{{- with .Values.ingress.annotations }}
  annotations:
    {{- . | toYaml | nindent 4 }}
{{- end }}
```

src/_base/helm/app/templates/application/webapp/ingress.yaml
```diff
@@ -3,9 +3,9 @@
...
 kind: Ingress
 metadata:
-{{- if .ingress.annotations }}
+{{- if $.Values.ingress.annotations }}
   annotations:
-  {{- range $key, $value := .ingress.annotations }}
+  {{- range $key, $value := $.Values.ingress.annotations }}
     {{ $key }}: {{ $value | quote }}
   {{- end }}
 {{- end }}
```
andytson-inviqa: similar here though I can't do a suggestion for unedited lines

src/_base/helm/app/values.yaml.twig
```diff
@@ -3,8 +3,11 @@ appVersion: {{ @('app.version') | json_encode | raw }}
 feature: {{ to_nice_yaml(@('helm.feature'), 2, 2) | raw }}
 
 ingress:
+  annotations: {{ to_nice_yaml(
+    (@('pipeline.base.istio.enabled') ? @('pipeline.base.istio.ingress.annotations') : @('pipeline.base.services.nginx.ingress.annotations')),
+    2, 4) | raw }}
```
andytson-inviqa: why would we need to specify both sets of annotations conditionally? I can only see if our harness was to set some up ourselves

kierenevans: Whilst I doubt the istio ingress controller would complain, an annotation like below doesn't make sense for istio's ingress:
```
nginx.ingress.kubernetes.io/proxy-body-size: 100m
```

andytson-inviqa: sorry if not clear what I mean, why not `@('pipeline.base.ingress.annotations')`

andytson-inviqa: do we set that in the harness, or are we considering it for the future to?

kierenevans: The annotations are customisable per project/per cluster and empty by default

andytson-inviqa: ok, so per project, where we are using both for different environments, and would be setting these in workspace.yml, rather than the cluster repo?

kierenevans: I guess it wouldn't have both on a project - it would be too inconsistent between environments. Will simplify tomorrow :+1:

kierenevans: Updated :)

src/_base/harness/attributes/docker-base.yml
```diff
@@ -221,7 +219,10 @@ attributes:
         tideways:
           environment:
             TIDEWAYS_HOSTNAME: = @('pipeline.base.hostname')
+      ingress:
+        annotations: {}
```
andytson-inviqa, line 223:
```suggestion
        annotations: {}
        type: standard # standard or istio
```

src/_base/harness/attributes/docker-base.yml
```diff
@@ -221,7 +219,10 @@ attributes:
...
       istio:
+        enabled: false
```
andytson-inviqa:
```suggestion
```

src/_base/helm/app/values.yaml.twig
```diff
@@ -3,8 +3,9 @@ appVersion: {{ @('app.version') | json_encode | raw }}
 feature: {{ to_nice_yaml(@('helm.feature'), 2, 2) | raw }}
 
 ingress:
+  annotations: {{ to_nice_yaml(@('pipeline.base.ingress.annotations'), 2, 4) | raw }}
   target_service: {{ @('services.varnish.enabled') ? 'varnish' : 'webapp' }}
-  type: "standard" # standard or istio
+  type: {{ @('pipeline.base.istio.enabled') ? 'istio' : 'standard' }}
```
andytson-inviqa:
```suggestion
  type: {{ @('pipeline.base.ingress.type') }}
```
since we already have it in a project and go harness (tho app.ingress latter), and the keys follow the same pattern in values and workspace

andytson-inviqa: could also go the whole length and do the full ingress as well by migrating the target_service to workspace.yml so it'd be
```
ingress: {{ to_nice_yaml(@('pipeline.base.ingress'), 2, 2) | raw }}
```

kierenevans: Done :+1:

andytson-inviqa (approved):

PR #547, opened 2021-02-03
Add Zed API hosts for supporting back-channel communication

andytson-inviqa:
through private DNS

Previous behaviour can be achieved by setting the api hosts as empty strings


kierenevans (dismissed):

kierenevans:
Maybe needs https://github.com/inviqa/harness-base-php/blob/1.1.x/src/_base/application/skeleton/README.md.twig#L43-L56 adjusting too?

kierenevans (dismissed):

kierenevans (approved):

PR #548, opened 2021-02-03
Implement ws external-images ls/rm and use for ws destroy --all

andytson-inviqa:
In order to allow developers to clear up as much space as they can per project when they've done with it.

It will complain if an image is already directly used by another project's container, but if it's not the direct commit but a FROM .., will just de-tag the image commit, keeping it around.


andytson-inviqa:
`ws destroy --all` doesn't actually seem to delete the external images, just services with both image: AND :build

kierenevans (approved):

PR #549, opened 2021-02-03
Add Jaeger Relay

kierenevans:
Forwards Jaeger traces to a global Jaeger instance provided by https://github.com/my127/workspace/pull/66 without having to put php-fpm and console in the shared my127 network and run into issues with docker-compose's DNS round robin for "mysql" or similar.

### DIFF
--- a/src/_base/docker/image/relay/root/etc/nginx/nginx.conf
+++ b/src/_base/docker/image/relay/root/etc/nginx/nginx.conf
@@ -13,7 +13,7 @@ stream {
   server {
     listen 1025;
 
-    # Avoid crashes on boot if jaeger isn't running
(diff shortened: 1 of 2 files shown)


src/_base/docker/image/relay/root/etc/nginx/nginx.conf
```diff
@@ -13,12 +13,50 @@ stream {
   server {
     listen 1025;
 
-    proxy_pass mailhog:1025;
+    # Avoid crashes on boot if jaeger isn't running
```
kierenevans:
```suggestion
    # Avoid crashes on boot if mailhog isn't running
```

andytson-inviqa (approved):

PR #550, opened 2021-02-03
Simplify cleanup in ws destroy

andytson-inviqa:
* only clean actually built images
* prune accepts build labels, so no need for xargs
* if buildkit is enabled then builder prune the label


kierenevans (approved):

kierenevans (changes requested):
Seems to clean external images too

kierenevans (approved):

PR #556, opened 2021-02-09
[DEVOPS-1181] Validating Lighthouse scores against configured pass values

g-foster2020:


src/_base/docker/image/lighthouse/root/app/run.sh.twig
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+
+lighthouse --no-enable-error-reporting \
+           --chrome-flags="--headless --no-sandbox=true" \
+           --output json \
+           --output-path=/home/headless/lighthouse-results.json \
+           ${TARGET_URL}
```
kierenevans: Fix from shellcheck:
```suggestion
           "${TARGET_URL}"
```

src/_base/harness/attributes/common.yml
```diff
@@ -156,6 +156,22 @@ attributes.default:
...
+        score: 0
+      best-practices:
+        enabled: 0
+        score: 0
+      accessibility:
+        enabled: 0
+        score: 0
+      performance:
+        enabled: 0
+        score: 0
```
kierenevans: For consistency with other true/false values:
```suggestion
    success-thresholds:
      pwa:
        enabled: false
        score: 0
      seo:
        enabled: false
        score: 0
      best-practices:
        enabled: false
        score: 0
      accessibility:
        enabled: false
        score: 0
      performance:
        enabled: false
        score: 0
```

src/_base/docker/image/lighthouse/Dockerfile.twig
```diff
@@ -3,7 +3,10 @@ FROM yukinying/chrome-headless-browser:latest
 USER root
 
 RUN apt update
+RUN apt install -y jq
+RUN apt install -y bc
 RUN apt install -y npm
```
kierenevans: The image doesn't need to be the smallest in final size or layers due to not being pushed anywhere, but it would still be a small speed improvement to work out the dependencies for all three packages in one go:
```suggestion
RUN apt update \
 && apt install -y bc jq npm
```

src/_base/harness/config/confd.yml
```diff
@@ -18,6 +18,7 @@ confd('harness:/'):
   - { src: docker/image/console/root/usr/local/etc/php/conf.d/docker-php-ext-tideways.ini }
   - { src: docker/image/console/root/usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini }
   - { src: docker/image/lighthouse/Dockerfile }
+  - { src: docker/image/lighthouse/root/app/run.sh }
```
kierenevans, line 21: This will need replicating into the `src/akeneo/harness/config/confd.yml` and the rest of the harness `confd.yml`s.

kierenevans (dismissed):
The confd.yml comment and shellcheck fix are needed and the rest are optional.

Could we have a Jenkinsfile.twig addition to run these tests by default, even if it wouldn't fail the build? There's one in _base and one in drupal8 so would need to be in both.

I'd also like to see this tested in this repository's `./test` script like we do for the other Jenkins tests: https://github.com/inviqa/harness-base-php/blob/1.1.x/test#L88
This will allow us to hopefully not break project's use of the lighthouse testing.

g-foster2020:
> The confd.yml comment and shellcheck fix are needed and the rest are optional.
> 
> Could we have a Jenkinsfile.twig addition to run these tests by default, even if it wouldn't fail the build? There's one in _base and one in drupal8 so would need to be in both.
> 
> I'd also like to see this tested in this repository's `./test` script like we do for the other Jenkins tests: https://github.com/inviqa/harness-base-php/blob/1.1.x/test#L88
> This will allow us to hopefully not break project's use of the lighthouse testing.

Great feedback as ever thanks. I will try to remember shellcheck from now on and the confd overrides.

I will implement all of these and your above suggestions.

kierenevans (approved):

PR #557, opened 2021-02-10
Prepare console build as early as possible

andytson-inviqa:
allows developers to start seeing the full codebase on the host as soon as it's available after build and introduces a new command just to get to that state `ws enable console`

### DIFF
--- a/src/_base/harness/scripts/enable.sh.twig
+++ b/src/_base/harness/scripts/enable.sh.twig
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-main()
+enable_all()
 {
     local IS_BUILT=no
     local HAS_FLAG=no
@@ -37,6 +37,12 @@ main()
         passthru docker-compose up -d
         passthru docker-compose exec -T -u build console app welcome
     fi
+}
+
+enable_console() {
+    passthru ws networks external
+    passthru ws external-images pull console
+    "${APP_BUILD}_console"
 }
 
 dynamic_console()
@@ -85,6 +91,11 @@ dynamic()
     {% endif %}
 }
 
+static_console()
+{
+    passthru docker-compose up --build -d console
+}
+
 static()
 {
     ws app build
@@ -102,4 +113,4 @@ initial_start()
     {% endif %}
 }
 
-main
+"enable_$1"
--- a/src/_base/harness/scripts/enable.sh.twig
+++ b/src/_base/harness/scripts/enable.sh.twig
@@ -25,6 +25,9 @@ main()
         fi
 
         "$APP_BUILD"
+
+        initial_start
+
         touch .my127ws/.flag-built
     else
         if [ "$APP_BUILD" = dynamic ] && [ "$USE_MUTAGEN" = yes ]; then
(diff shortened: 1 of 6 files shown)


src/_base/harness/scripts/enable.sh.twig
```diff
@@ -15,34 +15,40 @@ main()
...
+        IS_SYNCED=yes
+    fi
+
     # we synchronise then stop mutagen to avoid impacting CPU usage during the build
 
-    if [[ "$USE_MUTAGEN" = "yes" ]]; then
-        passthru ws mutagen start
+    if [ "$USE_MUTAGEN" = yes ]; then
+        if [ "$IS_SYNCED" = no ]; then
+            passthru ws mutagen start
```
andytson-inviqa: could do with renaming start as it's destructive, so `recreate` or making it an option `... start --recreate`

andytson-inviqa:
> allows developers to start seeing the full codebase on the host as soon as it's available after build

aka composer has finished install, do I need to wait for the rest of the init/startup process to open the codebase in an IDE? (a bigger problem also on `ws create` for spryker, where the demoshop is only in the container)

> unnecessary repeated deletion and recreation of the mutagen sync volume on errors at first startup

aka say jenkins fails to start, or mysql isn't available, why do I need to wait 5 minutes again to recreate and flush the project volume because of that?

andytson-inviqa:
clicked the wrong button :)

kierenevans (dismissed):

kierenevans:
Do we even need to pause and resume mutagen at all during setup anymore?
When we removed docker-sync in #530, we just re-worded the comment to not mention docker-sync as we didn't want to change too much.

There is definitely some CPU overhead to having mutagen syncing file changes to the host as they happen inside the container, but perhaps more modern macbooks are better at it.

If removing the pause/resume, need to check that the sync container actually exists or we will reintroduce a similar problem to #67.

andytson-inviqa:
>Do we even need to pause and resume mutagen at all during setup anymore?

I'm not sure, but also not confident to try it out just on the project I'm on when it might more affect magento2.

> perhaps more modern macbooks are better at it.

My laptop's the same as it was back then :)


Latest change I've done has changed this even further, as I realised while I was waiting for a Go dependency and some other things to build sequentially before I was able to consider the volume safe from wiping. On the plus side the workflow for the dynamic and static builds look much similar now, just that the `app build` is done on started container vs in the image build process

kierenevans (dismissed):

andytson-inviqa:
going to rethink this a little now we've got a PR targeting the sync volume issue itself. I'd prefer not to introduce another flag file but still can have console started earlier. Just perhaps need ws enable to allow more graceful re-runs before fully built than to throw everything away

src/_base/harness/scripts/enable.sh.twig
```diff
@@ -1,56 +1,98 @@
...
+
     passthru ws networks external
-    if [ "$IS_BUILT" = "no" ]; then
-        if [ "$HAS_FLAG" = "no" ]; then
-          passthru docker-compose down
+    if [ "$IS_BUILT" = no ]; then
+        if [ "$HAS_FLAG" = no ]; then
+            # remove all services, keeping console if considered built
+            if [ -f .my127ws/.flag-console-built ]; then
+                if [ -n "$(grep -v -Fxe console <<< "$SERVICES_CREATED")" ]
```
kierenevans:
```suggestion
                if [ -n "$(grep -v -Fxe console <<< "$SERVICES_CREATED")" ]; then
```

andytson-inviqa:
hmm, something's still stopping console still

edit: ah I see only dynamic was using the flag. static doesn't need it as the container created is enough to know it's built)

kierenevans (approved):

PR #558, opened 2021-02-10
Deprecate installing harness tools in php moving upstream

andytson-inviqa:
and avoid repeating installation whilst deprecated

### DIFF
--- a/src/_base/docker/image/php-fpm/Dockerfile.twig
+++ b/src/_base/docker/image/php-fpm/Dockerfile.twig
@@ -3,8 +3,8 @@ FROM {{ @('docker.repository') ~ ':' ~ @('app.version') }}-console as console
 LABEL build="{{ @('namespace') ~ ':' ~ @('app.version') }}"
 RUN if [ -d /app/tools/assets/ ]; then rm -rf /app/tools/assets/; fi
 RUN if [ -d {{ @('frontend.path') }}/node_modules/ ]; then rm -rf {{ @('frontend.path') }}/node_modules/; fi
-{% endif %}
 
+{% endif %}
 FROM {{ @('docker.image.php-fpm') }}
 WORKDIR /app
 COPY root  /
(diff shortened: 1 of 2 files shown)


andytson-inviqa:
the redis-cli seems to be unreliable as well, so I'll have a think about that (perhaps, if we get it working, we put the latest redis-cli release in the upstream image if it's backwards compatible with older servers)

src/_base/docker/image/php-fpm/Dockerfile.twig
```diff
@@ -9,13 +9,15 @@ FROM {{ @('docker.image.php-fpm') }}
...
+RUN ([ -e /sbin/tini ] || curl --fail --silent --location --output /sbin/tini https://github.com/krallin/tini/releases/download/v0.19.0/tini) \
  && chmod +x /sbin/tini \
- && curl --fail --silent --location --output /usr/local/bin/mhsendmail https://github.com/mailhog/mhsendmail/releases/download/v0.2.0/mhsendmail_linux_amd64 \
+ && ([ -e /usr/local/bin/mhsendmail ] || curl --fail --silent --location --output /usr/local/bin/mhsendmail https://github.com/mailhog/mhsendmail/releases/download/v0.2.0/mhsendmail_linux_amd64) \
  && chmod +x /usr/local/bin/mhsendmail
+
 {%- set install_extensions=@('php.install_extensions')|merge(@('php.fpm.install_extensions'))|filter(v => v is not empty) %}
-{%- if install_extensions %} \
- && cd /root/installer \
+{%- if install_extensions %}
```
kierenevans:
```suggestion
{%- if install_extensions %} \
```
Should fix the build error - this gets put on the same line as `chmod +x /usr/local/bin/mhsendmail` by twig.

Perhaps there's a better way to do it?

kierenevans:
Is redis-cli an issue due to using an alpine redis image with a debian-based php image?
Latest redis sounds good though.

src/_base/docker/image/php-fpm/Dockerfile.twig
andytson-inviqa: The better way is to not use - in twig as it already does trim a bit anyway. I've removed most of them used there, and added one. makes xdebug the line after extensions rather than having 2 lines gap if <3, and still has an empty line above it if no extensions

kierenevans (dismissed):

kierenevans (approved):

PR #560, opened 2021-02-12
Add missing translator cache and run split codebucket store tasks

andytson-inviqa:
* `translator:generate-cache` is needed for Zed's translations
* `stores: true` is needed for caches that split by store codebucket

### DIFF
--- a/src/spryker/application/skeleton/config/install/docker.yml.twig
+++ b/src/spryker/application/skeleton/config/install/docker.yml.twig
@@ -94,12 +94,18 @@ sections:
     cache:
         router-cache-warmup-yves:
             command: "vendor/bin/yves router:cache:warm-up"
+            stores: true
 
         router-cache-warmup-zed:
             command: "vendor/bin/console router:cache:warm-up"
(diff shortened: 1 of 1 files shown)


src/spryker/application/skeleton/config/install/docker.yml.twig
```diff
@@ -94,12 +94,18 @@ sections:
...
         router-cache-warmup-zed:
             command: "vendor/bin/console router:cache:warm-up"
+            stores: true
 
         twig-cache-warmup:
             command: "vendor/bin/console twig:cache:warmer"
+            stores: true
+
+        translator-generate-cache:
+            command: "vendor/bin/console translator:generate-cache"
```
hgajjar, line 108: good catch, it's even missing in Spryker demoshop installation recipe.

hgajjar (approved):

andytson-inviqa:
This doesn't fix all the issues (twig cache warmer not warming all twig templates), but still is an improvement

PR #561, opened 2021-02-12
Early feedback in jenkins by running minimal build for isolated tests

andytson-inviqa:
This is how we used to do it

### DIFF
--- a/src/_base/application/overlay/Jenkinsfile.twig
+++ b/src/_base/application/overlay/Jenkinsfile.twig
@@ -37,7 +37,7 @@ pipeline {
 {% if not @('jenkins.tests.isolated') %}
                 stage('acceptance') { steps { sh 'ws exec composer test-acceptance' } }
 {% endif %}
-                stage('helm kubeval qa')  { steps { sh 'ws helm kubeval qa' } }
+                stage('helm kubeval qa')  { steps { sh 'ws helm kubeval --cleanup qa' } }
                 stage('lighthouse') { steps { sh 'ws lighthouse' } }
             }
         }
--- a/src/_base/application/overlay/Jenkinsfile.twig
+++ b/src/_base/application/overlay/Jenkinsfile.twig
@@ -45,7 +45,6 @@ pipeline {
         stage('Build') {
             steps {
                 sh 'ws enable'
-                sh 'ws exec app composer:development_dependencies'
                 milestone(30)
             }
         }
--- a/src/_base/application/overlay/Jenkinsfile.twig
+++ b/src/_base/application/overlay/Jenkinsfile.twig
@@ -49,14 +49,10 @@ pipeline {
                 milestone(30)
             }
         }
-        stage('Acceptance') {
-            steps {
-                sh 'ws exec composer test-acceptance'
-            }
-        }
-        stage('Lighthouse') {
-            steps {
-                sh 'ws lighthouse'
+        stage('End-to-end Test') {
+            parallel {
+                stage('acceptance') { steps { sh 'ws exec composer test-acceptance' } }
+                stage('lighthouse') { steps { sh 'ws lighthouse' } }
             }
         }
 {% endif %}
--- a/src/_base/application/overlay/Jenkinsfile.twig
+++ b/src/_base/application/overlay/Jenkinsfile.twig
@@ -36,9 +36,9 @@ pipeline {
                 stage('unit')       { steps { sh 'ws exec composer test-unit'       } }
 {% if not @('jenkins.tests.isolated') %}
                 stage('acceptance') { steps { sh 'ws exec composer test-acceptance' } }
+                stage('lighthouse') { steps { sh 'ws lighthouse' } }
 {% endif %}
                 stage('helm kubeval qa')  { steps { sh 'ws helm kubeval --cleanup qa' } }
-                stage('lighthouse') { steps { sh 'ws lighthouse' } }
             }
         }
 {% if @('jenkins.tests.isolated') %}
@@ -52,6 +52,11 @@ pipeline {
         stage('Acceptance') {
             steps {
                 sh 'ws exec composer test-acceptance'
+            }
+        }
+        stage('Lighthouse') {
+            steps {
+                sh 'ws lighthouse'
             }
         }
 {% endif %}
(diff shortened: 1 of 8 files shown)


andytson-inviqa:
sorry, no starting console yet, but open to discussion

src/_base/application/overlay/Jenkinsfile.twig
```diff
@@ -6,9 +6,15 @@ pipeline {
...
+        stage('Setup') {
             steps {
-                sh 'ws install'
+                // ws install without enabling
+                sh 'ws install --step download'
+                sh 'ws install --step overlay'
+                sh 'ws install --step validate'
+                sh 'ws install --step prepare'
+
+                sh 'ws app build console'
```
andytson-inviqa: need to start console image for ws exec to work

kierenevans:
I really like this idea.

I'm a bit worried about magento2 needing to talk to redis and the database for things, but we can give it a go

Let's update our `test` script to replicate the new flow so we can see if it's working in this PR's build

src/_base/application/overlay/Jenkinsfile.twig
```diff
@@ -6,9 +6,15 @@ pipeline {
...
     stages {
-        stage('Build') {
+        stage('Setup') {
             steps {
-                sh 'ws install'
+                // ws install without enabling
+                sh 'ws install --step download'
+                sh 'ws install --step overlay'
+                sh 'ws install --step validate'
+                sh 'ws install --step prepare'
```
hgajjar: could we rather use `ws harness download` and `ws harness prepare` instead of individual steps? or does that not cover all of these steps?

andytson-inviqa:
seems the "if not fully built then blow the scaffolding down" currently prevents the composer dev dependencies from still being there for acceptance tests

test
```diff
@@ -80,28 +92,47 @@ function check_environment_started()
...
+    cd "${path_test}"
 
     # Check if we have a composer.json so we can run tests
     # Check inside the console container as static builds do not
     # update the local filesystem
     if ws exec stat /app/composer.json > /dev/null 2>&1; then
-      # Run checks that don't reqiure development dependencies
-      ws exec composer test-production-quality
       # Bring in dev dependencies
       ws exec app composer:development_dependencies
```
andytson-inviqa: I'd live with this if composer cache persisted between both runs

The alternative would be to somehow remove the need for the flag-built flag to avoid `docker-compose down` in enable.sh, but I expect it be complicated

kierenevans (changes requested):
I approve of the code, but the build seems to take an extra 20 minutes.
Even "php" and "akeneo" which don't use the new method are affected by up to a minute each

andytson-inviqa:
>Even "php" and "akeneo" which don't use the new method are affected by up to a minute each

don't you mean drupal/m1/m2, as their the ones that don't use the early feedback way

andytson-inviqa:
the composer install is the biggest contributor to slowdown. I've updated to stop it doing it in this PR twice for drupal/m1/m2, but as commented will happen for the isolated tests mode twice still due to the recreation of the console on `ws enable`

next will be that because unit/quality tests aren't run in parallel with acceptance there may be loss of overall time in order to bring the feedback earlier

andytson-inviqa:
I'm not sure I can do anything about the extra composer install for the moment. A separate PR later for making ws enable be more granularly graceful (avoiding stopping any services, not just console, that are already started, more like dc up -d) maybe

kierenevans:
I was wondering why it didn't just use docker image build cache and it's due to the `COPY . /app` step picking up these two files:
```bash
diff ../manifest-before.txt ../manifest-after.txt
93a94,95
> 6fa7dda84126f41fb3dbea18196e295042f214c3ef6b606fee542de6ac6c12d2  ./.my127ws/helm/qa/charts/ci-akeneo-sample-static-0.0.1.tgz
> e1264615f4c8a74c01f5b9671176568142a054b0e3a9e6a6c3fdae8027929310  ./.my127ws/helm/qa/requirements.lock
```
Cleaning up these two files (and the charts folder ittself) lets the console image re-use the cache.

test
```diff
@@ -80,27 +92,49 @@ function check_environment_started()
     fi
 )
 
+function run_isolated_tests()
+(
+  local harness="$1"
+  cd "${path_test}"
+
+  ws helm kubeval qa
+  rm -rf .my127ws/helm/qa/charts/ .my127ws/helm/qa/requirements.lock
```
kierenevans: This is better off in the `ws helm kubeval qa` method so we can fix in project Jenkins pipelines too

andytson-inviqa:
I was thinking more of the dev dependencies installed after the container, but it's all just theory I haven't checked

test
```diff
@@ -98,7 +98,6 @@ function run_isolated_tests()
   cd "${path_test}"
 
   ws helm kubeval qa
-  rm -rf .my127ws/helm/qa/charts/ .my127ws/helm/qa/requirements.lock
```
andytson-inviqa: I find it useful from a dev perspective keeping it, as I need to run it more than once

test
andytson-inviqa: I find it better from a dev perspective keeping it, as I usually need to run it more than once. We are dealing with yaml and all as well as its templating :)

kierenevans: Haha true, okay I'll made it an option

kierenevans:
Now on par or quicker than the main branch :+1: 
Once https://github.com/inviqa/harness-base-php/pull/557 is reworked a bit, would love to see this in 1.1.0!

src/_base/application/overlay/Jenkinsfile.twig
```diff
@@ -23,11 +34,32 @@ pipeline {
...
+        stage('Acceptance') {
+            steps {
+                sh 'ws exec composer test-acceptance'
             }
         }
+        stage('Lighthouse') {
+            steps {
+                sh 'ws lighthouse'
+            }
+        }
```
andytson-inviqa, line 56: can we run it in parallel with acceptance? I only made the acceptance one stage stage here because it was the only one

andytson-inviqa: maybe "End-to-end Test" the stage they are under

kierenevans: Sounds great :+1:

kierenevans (dismissed):

kierenevans (approved):

PR #565, opened 2021-02-18
Make curl fail helpfully and consistently

andytson-inviqa:
Using full option names to make it clear what they do, and use --location except for on unsupported request methods

### DIFF
--- a/src/_base/harness/scripts/mutagen.sh
+++ b/src/_base/harness/scripts/mutagen.sh
@@ -33,7 +33,7 @@ install_mutagen()
         return 1
     fi
 
-    run curl --fail --silent --show-error --location --disable --output .my127ws/utilities/mutagen/mutagen.tar.gz "$download_url"
+    run curl --fail --silent --show-error --location --output .my127ws/utilities/mutagen/mutagen.tar.gz "$download_url"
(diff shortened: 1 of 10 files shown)


src/_base/harness/scripts/mutagen.sh
```diff
@@ -33,7 +33,7 @@ install_mutagen()
         return 1
     fi
 
-    run curl -L -q -sS -f "$download_url" -o .my127ws/utilities/mutagen/mutagen.tar.gz
+    run curl --fail --silent --show-error --location --disable --output .my127ws/utilities/mutagen/mutagen.tar.gz "$download_url"
```
andytson-inviqa: I'm not sure -q was intentionally used to mean --disable (disables use of .curlrc), or mistaking for quiet?

hgajjar: Mistaken for quiet 🤦 
I like the full flag naming instead of abbreviations 👍

hgajjar (dismissed):
Looks good to me!

src/_base/harness/scripts/mutagen.sh
andytson-inviqa: yes, it's generally recommended to use long options for scripting, and only use short options for personal usage

kierenevans (approved):

PR #566, opened 2021-02-18
Deprecate istio sidecar shutdown on Job resources

andytson-inviqa:
istio will use k8s 1.18 sidecar lifecycle to terminate once other containers have finished (also fixes container startup without the proxy being ready as well)


andytson-inviqa:
nvm, it's still behind an api server feature flag so not in the k8s api yet

PR #568, opened 2021-03-03
Introduce a ws rebuild command

andytson-inviqa:
While `ws destroy; ws enable` can rebuild, it destroys the mutagen volume as well, which adds a lot of unnecessary time if the volume isn't 100% different to the project folder

This is in conjunction with #557, as ws enable will still destroy the mutagen volume without that

### DIFF
--- a/src/_base/harness/scripts/rebuild.sh
+++ b/src/_base/harness/scripts/rebuild.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+run docker-compose down --rmi local --volumes --remove-orphans --timeout 120
+
+passthru ws cleanup built-images
+
+run rm -f .my127ws/.flag-built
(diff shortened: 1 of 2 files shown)


kierenevans (approved):

src/_base/harness/scripts/rebuild.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+run docker-compose down --rmi local --volumes --remove-orphans --timeout 120
+
+passthru ws cleanup built-images
+
+run rm -f .my127ws/.flag-built
```
andytson-inviqa, line 7: just a thought, I probably would actually want to delete .flag-console-built as well, as that should be rebuilt, just not the mutagen volume

PR #572, opened 2021-03-04
Do not remove sync container, make more responsive to stop

kierenevans:
Do not remove the sync container during `ws mutagen start`.

Make `docker stop <syncContainer>` more responsive. The tail command seems to not handle the TERM signal that `docker stop` sends, so ends up waiting for 10 seconds to be KILLed by docker. `sleep` on it's own is also not responsive.

### DIFF
--- a/src/_base/harness/scripts/mutagen.sh
+++ b/src/_base/harness/scripts/mutagen.sh
@@ -52,7 +52,7 @@ setup_sync_container()
             done < <(echo "${VOLUME_MAPPINGS}" | grep "^${CONTAINER_NAME}" | cut -d ':' -f2- )
 
             # shellcheck disable=SC2046
-            passthru docker run -d --name "${CONTAINER_NAME}" $(printf -- '-v %q ' "${CONTAINER_VOLUME_MAPPINGS[@]}") alpine:latest tail -f /dev/null
+            passthru docker run -d --name "${CONTAINER_NAME}" $(printf -- '-v %q ' "${CONTAINER_VOLUME_MAPPINGS[@]}") alpine:latest /bin/sh -c "'trap : TERM INT; sleep 9999999999d & wait'"
         fi
     done
 }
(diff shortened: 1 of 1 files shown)


andytson-inviqa (dismissed):

src/_base/harness/scripts/mutagen.sh
```diff
@@ -43,17 +43,17 @@ setup_sync_container()
...
+        if [[ "$(docker ps --all --filter "name=${CONTAINER_NAME}\$" --format '{{.Names}}')" == "${CONTAINER_NAME}" ]]; then
+            passthru docker start "${CONTAINER_NAME}"
+        else
+            CONTAINER_VOLUME_MAPPINGS=()
+            while IFS= read -r line; do
+                CONTAINER_VOLUME_MAPPINGS+=("$line")
+            done < <(echo "${VOLUME_MAPPINGS}" | grep "^${CONTAINER_NAME}" | cut -d ':' -f2- )
+
+            # shellcheck disable=SC2046
+            passthru docker run -d --name "${CONTAINER_NAME}" $(printf -- '-v %q ' "${CONTAINER_VOLUME_MAPPINGS[@]}") alpine:latest /bin/sh -c "'trap : TERM INT; sleep 9999999999d & wait'"
```
andytson-inviqa: the "' '" is a bit odd, is it still a bug to do with workspace version? The pre-release would error for it

andytson-inviqa: seeing as we're soon getting that released (driven by php 8 update rather than our other changes), we prob need something compatible

kierenevans: Yes my `~/.my127/workspace/lib/sidekick.sh` is the old version at the moment.

andytson-inviqa: and I'm guessing compatible way is only via passthru "..." single argument form

andytson-inviqa:
```suggestion
            passthru docker run -d --init --name "${CONTAINER_NAME}" $(printf -- '-v %q ' "${CONTAINER_VOLUME_MAPPINGS[@]}") alpine:latest tail -f /dev/null
```
tini from docker to the rescue?

andytson-inviqa: (just not available from docker-compose v3 file format)

andytson-inviqa (approved):

PR #581, opened 2021-03-17
Move shellcheck to Jenkins. Add Hadolint

kierenevans:
* Remove travis
* Add shellcheck to quality checks in Jenkins
* Add hadolint to to quality checks in Jenkins
* Run quality checks on the prepared project in ./test
* Fixes for rendered shell scripts that fail shellcheck
* Fixes for existing hadolint issues


src/_base/docker/image/console/root/lib/task/build/backend.sh.twig
```diff
@@ -2,8 +2,11 @@
 
 function task_build_backend()
 {(
+    set -e -o pipefail
+
     cd {{ @('backend.path') }}
-    
+
+    # shellcheck disable=SC2160
     if [ ! {{ @('backend.build.when')|raw }} ]; then
```
andytson-inviqa, line 10: this seems to suggest that we'll have issues with when: "false" as I presume this was flagged because [ ! true ] rendered?

andytson-inviqa: in heinsight, incorporating the [ ] into the when prob would have been better

kierenevans: Yeah, it's as you suspect - `false` values get a complaint that you can just do `if ! false; then` instead of `if [ ! false ]; then`

andytson-inviqa:
@kierenevans let me know if you're ok with my change. It should expand out the arguments of an array into each for item

andytson-inviqa (approved):

kierenevans:
> @kierenevans let me know if you're ok with my change. It should expand out the arguments of an array into each for item

Yep, am happy

PR #585, opened 2021-03-26
Switch to msmtp to avoid DNS issues with mhsendmail

kierenevans:
@elvetemedve is having issues with mhsendmail resolving `mailhog-relay` incorrectly due to being compiled with an older go version and not picking up on docker's `options nodots:0` in resolv.conf

Using msmtp allows us to bypass this issue.

### DIFF
--- a/src/_base/docker/image/console/Dockerfile.twig
+++ b/src/_base/docker/image/console/Dockerfile.twig
@@ -7,8 +7,16 @@ RUN chown -R build:build /home/build /app \
  # installing tools in the image is deprecated
  && ([ -e /sbin/tini ] || curl --fail --silent --show-error --location --output /sbin/tini https://github.com/krallin/tini/releases/download/v0.19.0/tini) \
  && chmod +x /sbin/tini \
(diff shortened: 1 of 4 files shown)


src/_base/docker/image/console/Dockerfile.twig
```diff
@@ -7,8 +7,16 @@ RUN chown -R build:build /home/build /app \
...
+ && [ -e /usr/bin/msmtp ] || ( \
+    apt-get update -qq \
+    && DEBIAN_FRONTEND=noninteractive apt-get -qq -y --no-install-recommends install \
+    # package dependencies \
+      msmtp \
+    # clean \
+    && apt-get auto-remove -qq -y \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* \
+ ) \
```
andytson-inviqa: since the previous lines were changed to check only for BC, I don't think we need this installed here as it'd be in the upstream image

kierenevans: Okay let's merge https://github.com/my127/docker-php/pull/55 instead and I'll update this PR to remove it :+1:

kierenevans: Updated

andytson-inviqa (approved):

PR #586, opened 2021-03-31
Clean up old mutagen sync sessions

kierenevans:
We've seen a load average of 48 on a developer's machine due to multiple sync sessions. This reduced to less than 5 after cleaning up the old sync sessions.

This situation can arise if `mutagen.yml.lock` is removed (e.g. a clean checkout) and `ws mutagen start` or `ws install` is called on macOS.

`mutagen.yml.lock` contains a project ID that `mutagen project` uses to link to the sync and forward sessions.


hgajjar (approved):
👏

PR #588, opened 2021-04-06
Use built image for lighthouse to speed up environments

kierenevans:
Extracted the docker image build into https://github.com/inviqa/docker-lighthouse which is built to `quay.io/inviqa_images/lighthouse:latest`

### DIFF
--- a/src/_base/harness/config/commands.yml
+++ b/src/_base/harness/config/commands.yml
@@ -308,9 +308,10 @@ command('generate token <length>'):
     }
     echo $string;
 
-command('lighthouse'):
+command('lighthouse [--with-results]'):
   env:
     COMPOSE_PROJECT_NAME: = @('namespace')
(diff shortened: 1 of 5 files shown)


src/_base/harness/config/commands.yml
```diff
@@ -308,9 +308,10 @@ command('generate token <length>'):
     }
     echo $string;
 
-command('lighthouse'):
+command('lighthouse [--with-results]'):
```
andytson-inviqa, line 311: are the results of any use by default?

kierenevans: Perhaps, but it's a huge JSON object that could take up a lot of jenkins log disk space if output.

I guess we'd want it stored as a file and as a build artefact for referencing if we want this in the pipeline normally.

andytson-inviqa (approved):

PR #596, opened 2021-04-29
Consistent use of enabled param value in if statement

g-foster2020:


g-foster2020:
I'm not sure if this was left as it was a possibly breaking change...?

If not, I sampled some projects and they all use it as expected (enabled = yes). But I wanted to disable it so had to set it to false, without this change.

kierenevans (commented):
Unsure why it was left as a truthy comparison rather than an explicit one, but if it wasn't intentional then we will need to update the src/drupal8/application/overlay/Jenkinsfile.twig too

g-foster2020:
I updated the override. I can't find any project that sets it in workspace.yml as anything other than yes or no.

At the same time, in common.yml we set it as pipeline.publish.chart.enabled: no e.g. https://github.com/inviqa/harness-magento2/blob/1605f71acdd364fecbf455d0b86dd2a2ff6893b2/harness/attributes/common.yml

andytson-inviqa (changes requested):
not needed as yaml processes yes and true as boolean true rather than string "yes". If we want to avoid confusion we should change all yes to true instead

In environment variables is only where we use quoted "yes" (or in most case since symfony expressions are used, there is no yes equivalence to true and all strings have to be quoted anyway)

andytson-inviqa:
ah actually, my mistake, it's removed in Yaml 1.2. just I encounter it still in other tools as boolean

andytson-inviqa:
personally I prefer to use booleans though for booleans

andytson-inviqa (approved):

andytson-inviqa:
superseded by https://github.com/inviqa/harness-base-php/pull/613

PR #604, opened 2021-05-14
Use memory for SQLite testing in drupal8

kierenevans:
Via @opdavies 

Co-Authored-By: Oliver Davies <oliver.davies@inviqa.com>

### DIFF
--- a/src/drupal8/application/overlay/phpunit.xml.twig
+++ b/src/drupal8/application/overlay/phpunit.xml.twig
@@ -22,7 +22,7 @@
         <!-- Example SIMPLETEST_BASE_URL value: http://localhost -->
         <env name="SIMPLETEST_BASE_URL" value="https://{{ @('hostname') }}/"/>
         <!-- Example SIMPLETEST_DB value: mysql://username:password@localhost/databasename#table_prefix -->
-        <env name="SIMPLETEST_DB" value="sqlite://localhost//dev/shm/test.sqlite"/>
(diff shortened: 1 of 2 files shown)


src/drupal8/application/overlay/phpunit.xml.twig
```diff
@@ -22,8 +22,8 @@
         <!-- Example SIMPLETEST_BASE_URL value: http://localhost -->
         <env name="SIMPLETEST_BASE_URL" value="https://{{ @('hostname') }}/"/>
         <!-- Example SIMPLETEST_DB value: mysql://username:password@localhost/databasename#table_prefix -->
-        <env name="SIMPLETEST_DB" value="sqlite://ignored//tmp/test.sqlite"/>
+        <env name="SIMPLETEST_DB" value="sqlite://localhost//dev/shm/test.sqlite"/>
```
andytson-inviqa: /dev/shm in docker is only 64mb by default, will that be a problem? or instead mounting a tmpfs volume?

kierenevans: It seems to work okay :+1:

andytson-inviqa: to me it's not whether it works ok now, it's that the /dev/shm is small and shared between different purposes. If the db is loaded on a project with lots of data, or just more than currently tested with, it would interfere with the other purposes. That's why I'm thinking a tmpfs volume

andytson-inviqa: the shm is adjustable too also in docker-compose.yml tho

andytson-inviqa (approved):

PR #606, opened 2021-06-03
Fix issue with secrets not being available before app-migrate for Helm

andytson-inviqa:
app-migrate uses a pre-upgrade helm hook, which means a secret has to exist already in the deployment.

This doesn't affect app-init as that is a post-install hook, but a hook is needed in any case. Similar was already implemented with argocd sync-waves.

Due to hook deletion policies, this might remove the secret along with the sealedsecret prior to the new sealedsecret, however that is more ideal anyway, to allow deployments/jobs to require to wait for the secret to be created

### DIFF
--- a/src/_base/helm/app/templates/_base_helper.tpl
+++ b/src/_base/helm/app/templates/_base_helper.tpl
@@ -51,9 +51,11 @@ metadata:
   name: {{ .root.Values.resourcePrefix }}{{ .service_name }}
   labels:
     {{- include "chart.labels" .root | nindent 4 }}
+    helm
     app.kubernetes.io/component: {{ .component | default .service_name }}
   annotations:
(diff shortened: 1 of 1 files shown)


kierenevans:
What happens to existing pods using that secret if it gets deleted?

andytson-inviqa:
> What happens to existing pods using that secret if it gets deleted?

nothing for environment variables unless scaled/rescheduled, though probably the secret files get deleted on secrets mounted as volumes

andytson-inviqa:
I'm concerned with the way this might impact things, not affecting currently anything I think as secrets aren't used as volumes yet, but for Spryker I need to port over a configmap/secret volume for pipeline jwt keys, which can be impacted if using helm.

PR #621, opened 2021-09-21
MISC: Fix missing extensions.json file issue for akeneo

hgajjar:


a-ast (dismissed):

kierenevans:
Does this work okay even in akeneo v4?

hgajjar:
> Does this work okay even in akeneo v4?

good question, It was working ok without it for v4, so I would remove it for it 👍

kierenevans (approved):

PR #628, opened 2021-10-15
Use new external network syntax

kierenevans:
See deprecation notice on https://docs.docker.com/compose/compose-file/compose-file-v3/#external-1


iwright (approved):
👍

PR #634, opened 2021-11-05
Install the correct architecture tini for php images

andytson-inviqa:


andytson-inviqa:
I'm so glad `dpkg --print-architecture` is correct (even for armhf)

kierenevans (approved):

PR #635, opened 2021-11-10
Mount bash_history to persist it among containers

### LINKED ISSUE #627: Persist bash history for console between containers
To improve the developer experience for harness upgrades, store `/home/build/.bash_history` on a named volume, allowing history of commands to persist between console containers for a project

Our current named volumes are here - https://github.com/inviqa/harness-base-php/blob/1.2.x/src/_base/docker-compose.yml.twig#L24-L40 - and the console container can reference them here - https://github.com/inviqa/harness-base-php/blob/1.2.x/src/_base/_twig/docker-compose.yml/service/console.yml.twig#L13-L17

alexandrajulius:
This should solve the following issue
https://github.com/inviqa/harness-base-php/issues/627

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@
 /tests/
 /tmp-test*/
 .idea
-.bash_history
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 /tests/
 /tmp-test*/
 .idea
+.bash_history
--- a/src/_base/_twig/docker-compose.yml/service/console.yml.twig
+++ b/src/_base/_twig/docker-compose.yml/service/console.yml.twig
@@ -15,6 +15,7 @@
       - ./.my127ws/application:/home/build/application
       - ./.my127ws/docker/image/console/root/lib/task:/lib/task
       - ./.my127ws:/.my127ws
+      - ./.bash_history:/home/build/.bash_history
 {% else %}
     image: {{ @('services.console.image') }}
 {% endif %}
--- a/src/_base/_twig/docker-compose.yml/service/console.yml.twig
+++ b/src/_base/_twig/docker-compose.yml/service/console.yml.twig
@@ -15,7 +15,7 @@
       - ./.my127ws/application:/home/build/application
       - ./.my127ws/docker/image/console/root/lib/task:/lib/task
       - ./.my127ws:/.my127ws
-      - ./.my127ws/.bash_history:/home/build/.bash_history
+      - ./.bash_history:/home/build/.bash_history
 {% else %}
     image: {{ @('services.console.image') }}
 {% endif %}
--- a/src/_base/_twig/docker-compose.yml/service/console.yml.twig
+++ b/src/_base/_twig/docker-compose.yml/service/console.yml.twig
@@ -15,7 +15,7 @@
       - ./.my127ws/application:/home/build/application
       - ./.my127ws/docker/image/console/root/lib/task:/lib/task
       - ./.my127ws:/.my127ws
-      - ./.bash_history:/home/build/.bash_history
+      - ./.my127ws/.bash_history:/home/build/.bash_history
 {% else %}
     image: {{ @('services.console.image') }}
 {% endif %}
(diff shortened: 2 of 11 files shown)


src/_base/_twig/docker-compose.yml/service/console.yml.twig
```diff
@@ -15,6 +15,7 @@
       - ./.my127ws/application:/home/build/application
       - ./.my127ws/docker/image/console/root/lib/task:/lib/task
       - ./.my127ws:/.my127ws
+      - ./.bash_history:/home/build/.bash_history
```
kierenevans, line 18:
```suggestion
      - ./.my127ws/.bash_history:/home/build/.bash_history
```
As it's from the project root that this is referenced

kierenevans:
I realised now that with my advice from yesterday, this won't survive harness updates (`ws harness update existing` and `ws harness update fresh`).

Perhaps it should end up being stored in the project root and added to the various application/skeleton/.gitignore files?

alexandrajulius:
good point, we can certainly do that on Teufel, but where should i add it (empty .bash_history + .gitignore entry) to provide it in skeletons for new projects?

andytson-inviqa:
@alexandrajulius https://github.com/inviqa/harness-base-php/search?q=filename%3Agitignore

andytson-inviqa (changes requested):
still needs gitignoring

.gitignore
```diff
@@ -3,3 +3,4 @@
 /tests/
 /tmp-test/
 .idea
+.bash_history
```
andytson-inviqa: sorry, not that file, all the others in the skeleton directories

alexandrajulius: Ah nice :) I see where the skeletons are. Will revert this, and add to the other gitignores

andytson-inviqa: hmm we'll have another problem if the file has to exist, as tho the skeleton will be there on `ws create`, the file wont on any other person's clones

andytson-inviqa: I think you're right though, just we'd need to create the file some other way

andytson-inviqa: actually it wouldn't help on the `ws create` either.

It'd be something like the enable.sh running `touch .bash_history`

andytson-inviqa: otherwise if the file doesn't already exist, docker-compose would create a directory and bind that to the container, which isn't what we'd want

alexandrajulius: yes, that's what docker-compose does... i can add `touch .bash_history` to 
src/_base/harness/config/events.yml
```
after('harness.install'): |
  #!bash
  ws enable
```
but then the file does not survive a refresh/install? I think it's better than not having it tho.
Alternatively people would have to create that file manually in their project root. Which is weird i think

andytson-inviqa: I'd suggest https://github.com/inviqa/harness-base-php/blob/995a60cf25b378708c812ff8b926829c39d825c8/src/_base/harness/scripts/enable.sh.twig rather than the after event itself.

The file will survive in this way, as `touch` only creates the file if it doesn't exist

src/_base/harness/scripts/enable.sh.twig
```diff
@@ -47,6 +47,7 @@ enable_all()
         initial_start
 
         touch .my127ws/.flag-built
+        touch .bash_history
```
alexandrajulius: @andytson-inviqa is this the right place?

Chemaclass:
Hi! Any updates on this PR?

Chemaclass:
@kierenevans @andytson-inviqa is there any missing work to be done in this PR? I just wonder why it was postpone that much, because it's already half a year open without any new feedback 🤔

kierenevans:
@Chemaclass Unfortunately it's making the .bash_history as a directory via docker-compose:
```
➜  tmp-test git:(feature/persist-bash-history-among-containers) ls -la
total 336
drwxr-xr-x  9 kevans  1000   4096 May  6 10:04 .
drwxr-xr-x  7 kevans users   4096 May  6 10:03 ..
-rw-r--r--  1 kevans  1000    142 May  6 10:03 auth.json
drwxr-xr-x  2 root   root       6 May  6 10:03 .bash_history
```
I'll try to fix

PR #638, opened 2021-11-10
Enable bash history under spryker console

Chemaclass:
Original idea by @alexandrajulius 🙌

## 🛠 Manual steps
```
$ ws install
```
This will create an empty file .bash_history in `/.my127ws` and it will mount your local bash history to that file. 
Therefore, after exiting and entering the console container again, you will still have your bash history available.

## 🤔 Open questions

Maybe does it make sense to add this mounting to the _base console template? Or all other as well?
- `src/_base/_twig/docker-compose.yml/service/console.yml.twig` 

Answer: https://github.com/inviqa/harness-base-php/pull/635 ☝️

### DIFF
--- a/src/_base/harness/config/events.yml
+++ b/src/_base/harness/config/events.yml
@@ -2,6 +2,7 @@
 after('harness.install'): |
   #!bash
   ws enable
+  touch ./.my127ws/.bash_history
 
 after('harness.refresh'):
   env:
--- a/src/spryker/_twig/docker-compose.yml/service/console.yml.twig
+++ b/src/spryker/_twig/docker-compose.yml/service/console.yml.twig
@@ -16,6 +16,7 @@
       - ./.my127ws/application:/home/build/application
       - ./.my127ws/docker/image/console/root/lib/task:/lib/task
       - ./.my127ws:/.my127ws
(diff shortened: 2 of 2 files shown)


kierenevans:
Thank you for porting it from the other repo but https://github.com/inviqa/harness-base-php/pull/635 is already here :partying_face:

kierenevans:
Though actually, that PR doesn't cover the `src/spryker/_twig/docker-compose.yml/service/console.yml.twig` file, so both PRs will be needed or combined

Chemaclass:
@kierenevans jap, https://github.com/inviqa/harness-base-php/pull/635 is updating `_base` and this current PR is updating `spryker`.

alexandrajulius:
I wasn't aware, we need the mounting in `spryker` too? 
I tested the changes in [#635](https://github.com/inviqa/harness-base-php/pull/635) for only `base` and it works fine on `spryker`. 
I can add the change to #635 if this is required @kierenevans ?

kierenevans:
_base is the base layer, spryker is layered on top of that so need to apply to every `console.yml.twig` if there's an override in the framework-specific directories.

Chemaclass:
I close this PR in favor of the one from @alexandrajulius https://github.com/inviqa/harness-base-php/pull/635

PR #640, opened 2021-11-12
Update elasticsearch default to latest version (7.14.2)

andytson-inviqa:
only 7.8+ supports linux/arm64/v8

### DIFF
--- a/src/_base/harness/attributes/common.yml
+++ b/src/_base/harness/attributes/common.yml
@@ -310,7 +310,7 @@ attributes.default:
     host: elasticsearch
     image: elasticsearch
     port: 9200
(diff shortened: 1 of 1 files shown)


andytson-inviqa:
of note akeneo v4 says 7.5, v5 says 7.10, harness does 7.10.1
magento2 v2.3.7 says 7.9, v2.4.5 says 7.10, harness does 7.10.1 with https://github.com/inviqa/harness-base-php/pull/641
spryker just 6.x or 7.x

although if going with AWS then you'd be using opensearch instead (which has elasticsearch opendistro for older versions)

jameshalsall (approved):

PR #643, opened 2021-11-19
Switch mysql image to mysql/mysql-server if not amd64 host architecture, changing default to 8.0

andytson-inviqa:
So that projects will get a linux/arm64 version of mysql if using 8.0 by default.


andytson-inviqa:
What differences I could see in the Oracle official image were:

* base OS switch from debian to oraclelinux
* loss of support of *.sql.gz and *.sql.xz for compression of files injected into /docker-entrypoint-initdb.d/

kierenevans (approved):

PR #644, opened 2021-11-22
Add Spryker classmap for autoloading

jbarton123:
This builds up a classmap for Spryker and Spryker Shop to improve
the performance of autoloading on development and pipeline environments

### DIFF
--- a/src/spryker/application/skeleton/composer-harness.json.twig
+++ b/src/spryker/application/skeleton/composer-harness.json.twig
@@ -23,7 +23,11 @@
   "autoload-dev": {
     "psr-4": {
       "Acceptance\\": "features/bootstrap"
-    }
+    },
+    "classmap": [
+      "vendor/spryker",
+      "vendor/spryker-shop"
+    ]
   },
   "autoload": {
     "psr-4": {


src/spryker/application/skeleton/composer-harness.json.twig
```diff
@@ -23,7 +23,11 @@
   "autoload-dev": {
```
andytson-inviqa, line 23: as its in "autoload-dev", wouldn't that only improve local and CI, not pipeline environments, since typically pipeline skips "dev"?

jbarton123: You're right, it won't affect the environment - it's more for local and CI improvements 👍 

I can update the commit message against this to reflect that?

andytson-inviqa: ok to reword my question another way, would pipeline environments see any benefit in also having it? Or is it already in place in the sub packages?

jbarton123: I don't think pipeline environments would see a benefit (we haven't measured that on Atida, just that our CI jobs in Gitlab now take significantly less time to run).

andytson-inviqa: is the pipeline environment already using classmaps for these packages? If not then we should make it do so for both performance and consistency between environments

orukusaki: The full classmap is dumped for pipeline envs as part of the CI build

andytson-inviqa: ok thanks, I just checked and definitely can see vendor/spryker and vendor/spryker-shop in pipeline environments classmap

andytson-inviqa (approved):

PR #649, opened 2021-12-03
Add volumeMode/Name support to k8s pvcs

andytson-inviqa:
So existing PVs that don't have labels can be reclaimed if PVC was deleted


kierenevans:
Is `pick` a helm v3 feature only?

andytson-inviqa:
It actually comes from https://github.com/Masterminds/sprig, so might depend what version 2 uses. Helm doesn't document all of sprigs functionality

andytson-inviqa:
it templates without error tho, so I'm happy with it

kierenevans (approved):

PR #653, opened 2021-12-16
Upgrade all elasticsearch versions to latest versions

andytson-inviqa:
Even if a platform doesn't explicitly say it supports beyond 7.10, it will need upgrading

Spryker's elasticsearch 6 can use 6.8.22

### DIFF
--- a/src/akeneo/harness.yml
+++ b/src/akeneo/harness.yml
@@ -134,11 +134,6 @@ attributes:
     image:
       console: = 'my127/akeneo:' ~ @('php.version') ~ '-fpm-buster-console'
       php-fpm: = 'my127/akeneo:' ~ @('php.version') ~ '-fpm-buster'
-  elasticsearch:
-    image: elasticsearch
-    tag: '7.10.1'
-  mysql:
-    tag: '8.0'
   nginx:
     # used to limit what is copied into an Nginx static-built image
     # vendor directory needed for vendor/*/*/*/public/ resources (images etc.)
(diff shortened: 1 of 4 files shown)


src/akeneo/harness.yml
```diff
@@ -134,11 +134,6 @@ attributes:
     image:
       console: = 'my127/akeneo:' ~ @('php.version') ~ '-fpm-buster-console'
       php-fpm: = 'my127/akeneo:' ~ @('php.version') ~ '-fpm-buster'
-  elasticsearch:
-    image: elasticsearch
-    tag: '7.10.1'
-  mysql:
-    tag: '8.0'
```
kierenevans, line 141: huh?

andytson-inviqa: it's 8.0 in base

kierenevans: Ah is the default 8.0 now?

kierenevans (dismissed):

src/akeneo/harness.yml
andytson-inviqa: yep, so devs don't suffer the performance penalty of mysql/mysql-server:5.7 also not supporting arm64 (though it handles the platform choice better than mysql:5.7 insisting on platform being hard-coded

andytson-inviqa:
Update elasticsearch to 7.16.2 for additional CVE fix

andytson-inviqa:
Exclude upgrading Spryker elasticsearch

it will be done in #655

coledave (approved):

PR #658, opened 2022-01-21
Allow defining replicas from attributes for application deployments, …

andytson-inviqa:
…removing the existing setting by default

While its not relevant to scale console and cron beyond 1, it may be useful to scale to 0 without "enabled: false" removing the deployment and any related resources

### DIFF
--- a/src/_base/helm/app/templates/application/webapp/deployment.yaml
+++ b/src/_base/helm/app/templates/application/webapp/deployment.yaml
@@ -11,7 +11,9 @@ metadata:
   annotations:
     argocd.argoproj.io/sync-wave: "15"
 spec:
-  replicas: 1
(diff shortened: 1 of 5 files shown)


src/_base/helm/app/templates/application/webapp/deployment.yaml
```diff
@@ -11,7 +11,9 @@ metadata:
   annotations:
     argocd.argoproj.io/sync-wave: "15"
 spec:
-  replicas: 1
+  {{- with (pick . "replicas") }}
```
andytson-inviqa:
```suggestion
  {{- with (pick .Values.services.webapp "replicas") }}
```

kierenevans (approved):

PR #663, opened 2022-02-04
Fix lighthouse result forwarding

hgajjar:
There were two issues:
- It wasn't allowing `ws lighthouse --with-results` command to pass `OUTPUT_RESULTS` env var correctly
- This ws command was executing docker-compose run with TTY, which was causing it to combine stdout and stderr, which was making it impossible to forward the result json output to a file on host machine (eg. `ws lighthouse --with-results > result.json`

### DIFF
--- a/src/_base/_twig/docker-compose.yml/service/lighthouse.yml.twig
+++ b/src/_base/_twig/docker-compose.yml/service/lighthouse.yml.twig
@@ -4,6 +4,7 @@
     command: "/bin/true"
     environment:
       TARGET_URL: "{{ @('lighthouse.target.url') | raw }}"
+      OUTPUT_RESULTS: "${OUTPUT_RESULTS:-}"
 {% if @('app.build') == 'dynamic' %}
     volumes:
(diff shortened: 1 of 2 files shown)


andytson-inviqa (approved):

src/_base/_twig/docker-compose.yml/service/lighthouse.yml.twig
```diff
@@ -4,6 +4,7 @@
     command: "/bin/true"
     environment:
       TARGET_URL: "{{ @('lighthouse.target.url') | raw }}"
+      OUTPUT_RESULTS: "${OUTPUT_RESULTS:-}"
```
andytson-inviqa, line 7:
```suggestion
      OUTPUT_RESULTS: ~
```

means read this variable as it's name from the environment

andytson-inviqa: actually, perhaps better if it avoids error your way when not run from that ws command

PR #664, opened 2022-02-11
Move phpstan configs to skeleton

hgajjar:
Projects would need to overwrite this files very often, so it belongs to skeleton rather than overlay


kierenevans:
I agree with the phpstan-baseline.neon move, but not sure about the phpstan.neon move - all projects should have the same set of standards

Chemaclass:
> I agree with the phpstan-baseline.neon move, but not sure about the phpstan.neon move - all projects should have the same set of standards

I am not sure about that, @kierenevans. There might be projects where the `level` might be different, but especially the `ignoreErrors` could be totally different depending on the project, no? For example, I am working on a project where I definitely have errors that I want to ignore on purpose that are 100% for our project level.

hgajjar:
@kierenevans thanks for your review, but I also agree with @Chemaclass. Also, we keep it under skeleton in magento2 harness too - https://github.com/inviqa/harness-base-php/blob/1.3.x/src/magento2/application/skeleton/phpstan.neon
Would you agree with it?

kierenevans:
True, okay :+1:

kierenevans (approved):

PR #668, opened 2022-02-17
Add spryker refresh command

michaeltrestianu18:


src/spryker/harness.yml
```diff
@@ -234,6 +234,24 @@ command('frontend zed watch'):
...
+    passthru vendor/bin/console propel:database:drop
+    passthru vendor/bin/console propel:database:create
+    passthru vendor/bin/console propel:pg-sql-compat
+    passthru vendor/bin/console propel:migrate
+    passthru vendor/bin/console setup:init-db
+    passthru vendor/bin/console data:import
+    passthru vendor/bin/console cache:empty-all
+    passthru vendor/bin/console cache:class-resolver:build
+    passthru vendor/bin/console transfer:entity:generate
+    passthru vendor/bin/console rest-api:build-request-validation-cache
```
hgajjar: I don't think this would be much useful, as normally Spryker requires a lot many other things to reset the database and reinstall it. For ex, there is queue sync commands which syncs the data from postgres to redis/es which is missing here.
I would suggest to use the workspace's wrapper commands to achieve this. The following commands should be enough
```
passthru "ws exec vendor/bin/console propel:database:drop"
passthru "ws exec vendor/bin/console propel:database:create"
passthru "ws exec app install"
```
the `app install` command runs everything you need to reinstall Spryker after recreating the db. 
However, if the project is using kafka-outbox service, you would also need to restart the outbox container to let it create the outbox table after recreating postgres db (`passthru "docker-compose restart outbox-relay"`), tho we have a separate `ws app reinstall` command available at project level for this purpose.

hgajjar (approved):

PR #670, opened 2022-02-18
Fix issue with flex, with Symfony upgrade to 5.4 and sync-recipes on init

andytson-inviqa:


src/symfony/.ci/sample-static/workspace.yml
```diff
@@ -2,11 +2,11 @@ workspace('ci-symfony-sample-static'):
   description: generated local workspace for ci-symfony-sample.
   harness: inviqa/symfony
 
-attribute('app.build'): static
```
hgajjar: Should we add a comment here explaining why we are not setting the build to static here? For ex: "we are not setting the `app.build: static` here because we need to get repo artifacts by running it dynamically first, for more info check `init.sh`"

hgajjar (approved):

PR #674, opened 2022-03-25
Allow specifying k8s context on sealed-secret commands

andytson-inviqa:
and print the context on the output (after input due to blob's editor view losing visibility of the first message)

### DIFF
--- a/src/_base/harness/config/secrets.yml
+++ b/src/_base/harness/config/secrets.yml
@@ -1,17 +1,22 @@
-command('secret image-pull-config [--cert=<cert>] [--scope=<scope>] [--namespace=<namespace>]'):
+command('secret image-pull-config [--cert=<cert>] [--scope=<scope>] [--context=<context>] [--namespace=<namespace>]'):
   env:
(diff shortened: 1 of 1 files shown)


kierenevans (dismissed):

src/_base/harness/config/secrets.yml
```diff
@@ -1,19 +1,24 @@
...
   exec: |
     #!bash
-    if [ "$SEALED_SECRETS" == 'yes' ] && ! command -v kubeseal >/dev/null; then
-      echo 'kubeseal is needed in order to use this command' >&2
+    if [ "$SEALED_SECRETS" == 'yes' ] && ! command -v kubeseal kubectl >/dev/null; then
+      echo 'kubeseal and kubectl are needed in order to use this command' >&2
       exit 1
     fi
 
+    if [ -z "${K8S_CONTEXT:-" ]; then
```
andytson-inviqa:
```suggestion
    if [ -z "${K8S_CONTEXT:-}" ]; then
```

kierenevans (approved):

PR #679, opened 2022-04-21
Ignore null template key values

jbarton123:
Currently for values set to `~` that are compiled together using this function, these values are still appended to the output. This ensures that for any null values passed into this function, that they are ignored.

### DIFF
--- a/src/_base/_twig/docker-compose.yml/service/nginx.yml.twig
+++ b/src/_base/_twig/docker-compose.yml/service/nginx.yml.twig
@@ -3,7 +3,7 @@
 {% set syncvolume = true %}
 {% endif %}
 {% set hostnames = [@('hostname')] %}
-{% set hostnames = hostnames|merge(@('hostname_aliases')|map(alias => "#{alias}." ~ @('domain'))) %}
+{% set hostnames = hostnames|merge(@('hostname_aliases')| filter(v => v is not empty) | map(alias => "#{alias}." ~ @('domain'))) %}
 
   nginx:
     build: .my127ws/docker/image/nginx
--- a/src/_base/harness/config/functions.yml
+++ b/src/_base/harness/config/functions.yml
@@ -255,6 +255,9 @@ function('template_key_value', [template, key_value]): |
   #!php
   $output = [];
   foreach ($key_value as $key => $value) {
+    if (!$value) {
+      continue;
+    }
     $output[str_replace('{{key}}', $key, $template)] = $value;
   }
   = $output;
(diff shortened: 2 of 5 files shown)


src/_base/harness/config/functions.yml
```diff
@@ -255,6 +255,9 @@ function('template_key_value', [template, key_value]): |
   #!php
   $output = [];
   foreach ($key_value as $key => $value) {
+    if (!$value) {
```
andytson-inviqa:
```suggestion
    if ($value===null) {
```

andytson-inviqa: '0' is a valid string but !$value is true for it

hgajjar (approved):

src/spryker/docker/image/nginx/root/etc/nginx/conf.d/zed.conf.template.twig
```diff
@@ -85,4 +86,5 @@ server {
         fastcgi_param APPLICATION_STORE {{ store_code }};
     }
 }
+{% endif %}
```
andytson-inviqa, line 89: misses zed api further down, however there's another PR for that https://github.com/inviqa/harness-base-php/pull/685

andytson-inviqa (changes requested):

src/_base/_twig/docker-compose.yml/service/nginx.yml.twig
```diff
@@ -3,7 +3,7 @@
 {% set syncvolume = true %}
 {% endif %}
 {% set hostnames = [@('hostname')] %}
-{% set hostnames = hostnames|merge(@('hostname_aliases')|map(alias => "#{alias}." ~ @('domain'))) %}
+{% set hostnames = hostnames|merge(@('hostname_aliases')| filter(v => v is not empty) | map(alias => "#{alias}." ~ @('domain'))) %}
```
andytson-inviqa, line 6: Just this bit now that is what we want to keep after @Chemaclass's PRs now merged

andytson-inviqa: although should also be done for varnish. I'll create a PR

src/_base/harness/config/functions.yml
```diff
@@ -255,6 +255,9 @@ function('template_key_value', [template, key_value]): |
   #!php
   $output = [];
   foreach ($key_value as $key => $value) {
+    if ($value === null) {
+      continue;
+    }
```
Chemaclass, line 260: Already done: https://github.com/inviqa/harness-base-php/pull/686

src/_base/_twig/docker-compose.yml/service/nginx.yml.twig
andytson-inviqa: https://github.com/inviqa/harness-base-php/pull/689, I've instead done at the source

PR #686, opened 2022-04-28
Ignore null values from template_key_value func

Chemaclass:
## Description

Ignore all null values from the `key=>values`. 
Also add an early return `[]` if the `$key_value` is already empty anyhow.

### DIFF
--- a/src/_base/harness/config/functions.yml
+++ b/src/_base/harness/config/functions.yml
@@ -253,8 +253,8 @@ function('replace', [haystack, needle, replacement]): |
 
 function('template_key_value', [template, key_value]): |
   #!php
-  if ($key_value === null) {
-    return null;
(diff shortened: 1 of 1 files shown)


src/_base/harness/config/functions.yml
```diff
@@ -253,8 +253,14 @@ function('replace', [haystack, needle, replacement]): |
 
 function('template_key_value', [template, key_value]): |
   #!php
+  if ($key_value === null) {
+    return null;
```
andytson-inviqa: not sure if empty array is more appropriate

Chemaclass: Good idea. I will change it to this:
```php
if (empty($key_value)) {
    return [];
}
```

andytson-inviqa (approved):

PR #689, opened 2022-04-28
Filter out empty spryker hostnames

andytson-inviqa:
Since the only way in workspace to remove an item is to make it empty.


andytson-inviqa:
Since it was a spryker-only issue, I didn't want to pepper the other harnesses with the empty check, so I've made a function to allow doing it at the source

Chemaclass (approved):
Nice 👌

kierenevans (approved):

PR #690, opened 2022-04-29
Fix issue with spryker jenkins-runner running as root user

andytson-inviqa:


andytson-inviqa:
While `su` might not forward signals, tini is correctly forwarding to the jenkins-agent-start script and terminating.
Jenkins jobs from spryker appear to be running ok

kierenevans (dismissed):

src/spryker/docker/image/jenkins-runner/root/lib/task/jenkins/setup.sh
```diff
@@ -10,4 +10,8 @@ function task_jenkins_setup()
 
     # link php where it is expected by cron.conf
     ln -s /usr/local/bin/php /usr/bin/php
+
+    # setup www-data folders
+    mkdir /var/www/.jenkins /data
+    chown www-data:www-data /var/www/.jenkins /data
```
andytson-inviqa: actually I'll move this and the ln -s to Dockerfile

kierenevans (approved):

PR #700, opened 2022-05-10
Use .max_memory instead of raw-value in redis

Chemaclass:
## 📚 Description

Right now we are using a raw value that define the max memory for redis-server: `"1073742000"`.
We could use the `.max_memory` variable instead.

### DIFF
--- a/src/_base/helm/app/templates/service/redis/deployment.yaml
+++ b/src/_base/helm/app/templates/service/redis/deployment.yaml
@@ -31,7 +31,7 @@ spec:
       - args:
         - redis-server
         - --maxmemory
-        - "1073742000"
+        - {{ .max_memory | quote }}
         - --maxmemory-policy
         - allkeys-lru
         - --save


kierenevans:
Should it be related to `.resources.memory` that we pass into `.spec.resources.requests.memory`? 

The background save needs roughly double the amount of memory to be able to save to disk successfully, so maybe the algorithm `(.spec.resources.requests.memory in bytes / 2) - (some overhead value in case of multiple processes)`?

If we keep `.max_memory` as-is, let's define a default for it in the helm chart and attributes :+1:

`redis-session` and perhaps the docker-compose versions might need some attention as well

Chemaclass:
It sounds interesting what you're saying, but I am not sure I have the knowledge to do that right now. I pass it to you, so you can improve this area if you consider is important, @kierenevans :)

andytson-inviqa:
I think also, if using a new value, it should be `.options.maxmemory` like we define options on mysql.

andytson-inviqa:
this has been done in a more configurable way #728, although will be migrating to use https://github.com/inviqa/harness-docker/pull/85

PR #675, opened 2022-04-01
Update k8s Ingress to v1 and add optional ingressClassName

andytson-inviqa:
Needed for k8s 1.22, supported since 1.19, and requires ArgoCD > 1.8

PR #702, opened 2022-05-12
Add pathType to Spryker webapp ingress

Chemaclass:
## Description

While upgrading our project to the latest harness, I faced this error:
``` 
spec.rules[0].http.paths[0].pathType: Required value: pathType must be specified, spec.rules[0].http.paths[1].pathType: Required value: pathType must be specified
```

After some research, I came to this post https://pet2cattle.com/2021/04/ingress-pathtype where it seems to be missing this property:
```yaml
pathType: Prefix
```

The question that I have now is: which of the three possible options would be the optimised for this field?

### We have tree options for selecting the pathType
- ImplementationSpecific: Matching will be handled by the IngressClass. Depending on the implementations can treat this in a different way
- Exact: Matches the URL path exactly with case sensitivity.
- Prefix: Matches based on a URL path prefix split by / (most common) It is also case sensitive on a path element by element basis.

### DIFF
--- a/src/spryker/helm/app/templates/application/webapp/ingress.yaml
+++ b/src/spryker/helm/app/templates/application/webapp/ingress.yaml
@@ -20,7 +20,9 @@ spec:
   - host: {{ $value | quote }}
     http:
       paths:
-      - backend:
+      - path: /*
+        pathType: Prefix
+        backend:
           service:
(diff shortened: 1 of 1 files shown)


andytson-inviqa (approved):

andytson-inviqa:
Prefix is how we did the other ingresses https://github.com/inviqa/harness-base-php/pull/675/files

PR #704, opened 2022-05-13
Remove creationTimestamp from ingress

Chemaclass:
## 🔖 Changes

- Fix indentation for `with .Values.ingress.annotations` and `tls`
- Removed `creationTimestamp: null`

### DIFF
--- a/src/_base/helm/app/templates/application/webapp/ingress.yaml
+++ b/src/_base/helm/app/templates/application/webapp/ingress.yaml
@@ -2,11 +2,10 @@
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
-{{- with .Values.ingress.annotations }}
+  {{- with .Values.ingress.annotations }}
   annotations:
     {{- . | toYaml | nindent 4 }}
-{{- end }}
-  creationTimestamp: null
+  {{- end }}
   labels:
     {{- include "chart.labels" . | nindent 4 }}
     app.kubernetes.io/component: webapp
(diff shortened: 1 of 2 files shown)


andytson-inviqa:
I don't believe there's any issue here, all changes are where the whitespace is truncated `{{-`

andytson-inviqa:
though creationTimestamp you're right to remove (just as status shouldn't be in it)

Chemaclass:
ah, well, it make sense. I apply those indentation fixes because the IDE (PhpStorm) told me that so it looks nicer. Do you want me to rollback those changes? Or is it fine as it is right now? We could remove the creationTimestamp at least in this PR. What do you say, @andytson-inviqa?

andytson-inviqa:
If you rename the ticket to "Remove creationTimestamp from ingresses", then I'd be happy with the formatting changes still in there, however the same changes we'd also make in the base harness https://github.com/inviqa/harness-base-php/blob/1.4.x/src/_base/helm/app/templates/application/webapp/ingress.yaml

andytson-inviqa:
there's actually a few more places that have creationTimestamp in, so I can create a PR that does all of them

Chemaclass:
Yeah, there are creationTimestamp in Deployment, and StatefulSet, but not more Ingress:
Screenshot 2022-05-13 at 13 44 23
Are we sure we want to remove the creationTimestamp from those other places too?

andytson-inviqa:
I'll merge this PR and do a separate one of the other places

andytson-inviqa (approved):

andytson-inviqa:
You're right to bring up the question, as the rest are all in `template:` fields, it means that they'll all cause a rollout on release even if there's nothing else to change

PR #706, opened 2022-05-13
Use preview name for helm preview Chart

Chemaclass:
## 📚  Description

When running `helm lint` I got an error saying that the chart name should be the same as the directory where the Chart is located. Also, a warning saying that the `apiVersion` was missing.

### DIFF
--- a/src/spryker/helm/preview/Chart.yaml
+++ b/src/spryker/helm/preview/Chart.yaml
@@ -1,5 +1,6 @@
-name: app-preview
+name: preview
 description: Base helm chart for the app-preview environment
 version: 0.0.1
+apiVersion: v1
 sources:
 home:
--- a/src/spryker/helm/preview/Chart.yaml
+++ b/src/spryker/helm/preview/Chart.yaml
@@ -2,6 +2,5 @@ apiVersion: v1
 name: preview
 description: Base helm chart for the app-preview environment
 version: 0.0.1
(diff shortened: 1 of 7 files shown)


andytson-inviqa:
What version of helm are you using? I can't recreate the chart name issue on v3.8.2, but I'm aware this used to be an issue, which is why we never integrated helm lint, instead kubeval

andytson-inviqa:
What version of helm are you using? I can't recreate this issue on v3.8.2, but I'm aware this used to be an issue, which is why we never integrated helm lint, instead kubeval

Chemaclass:
I am using helm v2.17.0

src/spryker/helm/preview/Chart.yaml
```diff
@@ -1,6 +1,7 @@
 apiVersion: v1
-name: app-preview
+name: preview
 description: Base helm chart for the app-preview environment
 version: 0.0.1
+apiVersion: v1
```
andytson-inviqa: duplicated now

Chemaclass: ach, sorry! I resolved the conflict via the web browser... I will fix it from PhpStorm 😂

andytson-inviqa:
preview charts are no longer supported directly by the harness, however it wont stop projects which need to implement this. Additionally helm v2 support has been removed

PR #707, opened 2022-05-16
Add apiVersion to helm charts for lint error

andytson-inviqa:
While we still don't use `helm lint`, this solves the errors using it in helm 3.8.2


kierenevans (approved):

kierenevans:
A later build has passed so I think this is safe to merge. Unsure why github status was not updated though.

PR #710, opened 2022-05-18
Update permissions for kubectl.config.yaml

Chemaclass:
## Description

After upgrading to helm 3, I faced this issue and I found a potential solution for this too: https://github.com/helm/helm/issues/9115
```
WARNING: Kubernetes configuration file is group/world-readable 
```

### DIFF
--- a/src/_base/harness/config/pipeline.yml
+++ b/src/_base/harness/config/pipeline.yml
@@ -87,10 +87,7 @@ command('app deploy <environment>'):
     #!bash(harness:/helm)|=
     set -o pipefail
     cd "${ENVIRONMENT}"
-    doctl -t "$DO_ACCESS_TOKEN" kubernetes cluster kubeconfig show "$CLUSTER" > kubectl.config.yaml
-    if helm version --short --client | grep '^v3' >/dev/null 2>&1; then
-      chmod go-r kubectl.config.yaml
-    fi
+    (umask 0200 && doctl -t "$DO_ACCESS_TOKEN" kubernetes cluster kubeconfig show "$CLUSTER" > kubectl.config.yaml)
     if helm version --short --client | grep '^Client: v2' >/dev/null 2>&1; then
       passthru helm init --client-only
     fi
--- a/src/_base/harness/config/pipeline.yml
+++ b/src/_base/harness/config/pipeline.yml
@@ -87,7 +87,7 @@ command('app deploy <environment>'):
     #!bash(harness:/helm)|=
     set -o pipefail
     cd "${ENVIRONMENT}"
-    (umask 0200 && doctl -t "$DO_ACCESS_TOKEN" kubernetes cluster kubeconfig show "$CLUSTER" > kubectl.config.yaml)
+    (umask 0077 && doctl -t "$DO_ACCESS_TOKEN" kubernetes cluster kubeconfig show "$CLUSTER" > kubectl.config.yaml)
     if helm version --short --client | grep '^Client: v2' >/dev/null 2>&1; then
       passthru helm init --client-only
     fi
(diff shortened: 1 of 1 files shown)


src/_base/harness/config/pipeline.yml
```diff
@@ -88,6 +88,9 @@ command('app deploy <environment>'):
     set -o pipefail
     cd "${ENVIRONMENT}"
     doctl -t "$DO_ACCESS_TOKEN" kubernetes cluster kubeconfig show "$CLUSTER" > kubectl.config.yaml
+    if helm version --short --client | grep '^v3' >/dev/null 2>&1; then
+      chmod go-r kubectl.config.yaml
+    fi
```
andytson:
```suggestion
    (umask 0200 && doctl -t "$DO_ACCESS_TOKEN" kubernetes cluster kubeconfig show "$CLUSTER" > kubectl.config.yaml)
```

I think its fine making it for helm v2 as well. This makes it so the file never got go+r

andytson:
the reason we never saw this is because we started using ArgoCD (which we use helm template render) before the switch to helm v3, so this code was never later used with helm v3 except I think just by your project.

src/_base/harness/config/pipeline.yml
Chemaclass: Yeah, I think helm2 would be ok with this as well. Thanks! 🙏

src/_base/harness/config/pipeline.yml
```diff
@@ -87,7 +87,7 @@ command('app deploy <environment>'):
     #!bash(harness:/helm)|=
     set -o pipefail
     cd "${ENVIRONMENT}"
-    doctl -t "$DO_ACCESS_TOKEN" kubernetes cluster kubeconfig show "$CLUSTER" > kubectl.config.yaml
+    (umask 0200 && doctl -t "$DO_ACCESS_TOKEN" kubernetes cluster kubeconfig show "$CLUSTER" > kubectl.config.yaml)
```
andytson: silly me, wrong way around in my last suggestion
```suggestion
    (umask 0077 && doctl -t "$DO_ACCESS_TOKEN" kubernetes cluster kubeconfig show "$CLUSTER" > kubectl.config.yaml)
```

Chemaclass: Is this equivalent as the proposed solution from the issue: https://github.com/helm/helm/issues/9115#issuecomment-818832622 ?
> chmod 600 ~/.kube/config

andytson: yes, umask removes the bits that are allowed on a file on creation, so e.g. 0007 means if a file created with 666 permissions then it gets 660 permissions. You can see it easily with `(umask 0077 && touch testfile)` that it gets 600 permissions

src/_base/harness/config/pipeline.yml
```diff
@@ -87,7 +87,7 @@ command('app deploy <environment>'):
     #!bash(harness:/helm)|=
     set -o pipefail
     cd "${ENVIRONMENT}"
-    doctl -t "$DO_ACCESS_TOKEN" kubernetes cluster kubeconfig show "$CLUSTER" > kubectl.config.yaml
+    (umask 0077 && doctl -t "$DO_ACCESS_TOKEN" kubernetes cluster kubeconfig show "$CLUSTER" > kubectl.config.yaml)
```
Chemaclass, line 90: For the [record](https://github.com/inviqa/harness-base-php/pull/710#discussion_r876336860): `umask` removes the bits that are allowed on a file on creation, so e.g. 0007 means if a file created with 666 permissions then it gets 660 permissions. You can see it easily with (umask 0077 && touch testfile) that it gets 600 permissions.

PR #712, opened 2022-05-19
711 - Drupal: add default services.local.yml for local environment.

### LINKED ISSUE #711: Drupal: Add default services.local.yml for local environment.
At the moment we don't have local services turn on by default. This is needed for most development as it's the way to turn off Drupal caches, twig caches and more...

thomaslobjoie:


andytson-inviqa:
are .yml files excluded from web access?

andytson-inviqa:
also 'for local environment', isn't true, this file is used for all environments I believe? Just you yourself might not be using k8s

thomaslobjoie:
> 

Really? I thought settings.local.php was only loaded locally.

thomaslobjoie:
> are .yml files excluded from web access?

It doesn't look like it. Is that an issue?

andytson-inviqa:
I don't know, but there is only settings.local.* in the repo, and it's not specifically local settings

thomaslobjoie:
> > 
> 
> Really? I thought settings.local.php was only loaded locally.

On Phoenix, it looks like we changed the include like this:
```
if (isset($app_host) && $app_host === 'phoenix-drupal.my127.site') {
   include $app_root . '/' . $site_path . '/settings.local.php';
}
```
But in the harness it is:
```
if (file_exists($app_root . '/' . $site_path . '/settings.local.php')) {
  include $app_root . '/' . $site_path . '/settings.local.php';
}
```

andytson-inviqa:
so sounds like in harness it's more like the same as magento and sprkyer's "local" meaning, but your project more like local environment

PR #713, opened 2022-05-24
Create ns with kubectl when using helm3

Chemaclass:
## 📚  Description

When using helm3, the `app deploy preview` doesn't work because the namespace it's missing.

The solution is to create the k8s namespace right before running helm to ensure there is a k8s namespace in place.


andytson-inviqa (dismissed):

kierenevans (approved):

src/_base/harness/config/pipeline.yml
```diff
@@ -91,6 +91,9 @@ command('app deploy <environment>'):
     if helm version --short --client | grep '^Client: v2' >/dev/null 2>&1; then
       passthru helm init --client-only
     fi
+    if helm version --short --client | grep '^v3' >/dev/null 2>&1; then
+      kubectl --kubeconfig="${PWD}/kubectl.config.yaml" create ns "${NAMESPACE}" || true
+    fi
```
andytson, line 96: needs to be conditional for if the k8s server doesn't allow ns creation to the user. Not all uses will be necessarily for feature branch environments only, and so static environments may have namespace already created.

can't presume also can list namespaces either to check if it exists, although `kubectl get namespace "${NAMESPACE}"` might be useable.

andytson: ah, nvm I forgot I mentioned this before, and the || true stops it causing breakage

PR #716, opened 2022-06-09
Fix harness.yml to use /Backoffice as root_dir for spryker versions newer 202108.0

sebastianlarisch:
With Release 202108.0 Spryker has implemented separate application bootstrapping into individual endpoints.
The harness.yml was already updated to use /BackendGateway as the endpoint for the ZED API for Spryker demoshop version >= 202108.0, but the Zed endpoint is still pointing to /Zed. Following Sprykers convention it should be /Backoffice instead.

### DIFF
--- a/src/spryker/harness.yml
+++ b/src/spryker/harness.yml
@@ -123,7 +123,7 @@ attributes:
       DE: = 'zed-de-' ~ @('hostname')
       AT: = 'zed-at-' ~ @('hostname')
       US: = 'zed-us-' ~ @('hostname')
-    root_dir: = @('app.web_directory') ~ '/Zed'
+    root_dir: "= @('spryker.demoshop-version') >= '202108.0'? @('app.web_directory') ~ '/Backoffice' : @('app.web_directory') ~ '/Zed'"
   zed_api:
     external_hosts:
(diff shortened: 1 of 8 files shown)


hgajjar (approved):
👍

src/spryker/harness.yml
```diff
@@ -123,7 +123,7 @@ attributes:
       DE: = 'zed-de-' ~ @('hostname')
       AT: = 'zed-at-' ~ @('hostname')
       US: = 'zed-us-' ~ @('hostname')
-    root_dir: = @('app.web_directory') ~ '/Zed'
+    root_dir: "= @('spryker.demoshop-version') >= '202108.0'? @('app.web_directory') ~ '/Backoffice' : @('app.web_directory') ~ '/Zed'"
```
hgajjar: Actually, it's better to use the `version_compare` function here. Example: https://github.com/inviqa/harness-base-php/blob/1.5.x/src/spryker/harness.yml#L92

hgajjar (changes requested):

src/spryker/harness.yml
sebastianlarisch: k, changed it

Chemaclass:
Any update regarding this PR? 🎻

PR #717, opened 2022-06-17
Add the ability to list built images

andytson-inviqa:
e.g. for instance to scan them for vulnerabilities before push


andytson-inviqa:
moving to draft as I forgot this only was working for pipeline only

andytson-inviqa:
I'll stick with services with image for now (non local), as same as the cleanup

kierenevans (approved):

PR #723, opened 2022-07-11
Explicitly configure all composer plugins

andytson-inviqa:
It seems the behaviour doesn't match the documentation, instead failing install rather than just warning


andytson-inviqa:
https://github.com/composer/composer/issues/10928 regarding feedback on the behaviour

PR #739, opened 2023-01-19
Allow configuring the Akeneo v6+ messenger consumers

andytson-inviqa:
For instance the Events API needs `webhooks` if enabled


kierenevans (dismissed):

andytson-inviqa:
hmm, didn't trigger any akeneo builds

kierenevans (approved):

PR #740, opened 2023-01-19
Akeneo breakage bug fix

andytson-inviqa:


kierenevans (dismissed):

andytson-inviqa:
still has the following error even though I tried to skeleton the bundles.php

```
Kernel "Kernel" used by Behat with "test" environment and debug disabled needs to have "FriendsOfBehat\SymfonyExtension\Bundle\FriendsOfBehatSymfonyExtensionBundle" bundle registered.
```
behat runs fine if it's set.

g-foster2020 (approved):

PR #741, opened 2023-01-26
Fix console container startup issue

### LINKED ISSUE #666: Console container startup issue
For a sylius project (symfony harness), in enable.sh we have:
```bash
docker-compose up -d --build console
docker-compose exec -u build console app build
```

The frontend build that happens after composer install fails when there's nothing to download via composer:
```bash
■ > yarn install
Command failed. stderr:
Error: EACCES: permission denied, open '/home/build/.config/yarn'
    at Object.openSync (fs.js:443:3)
    at readFileSync (fs.js:343:35)
    at /home/build/.nvm/versions/node/v10.24.1/lib/node_modules/yarn/lib/cli.js:101466:58
    at Array.map (<anonymous>)
    at parseRcPaths (/home/build/.nvm/versions/node/v10.24.1/lib/node_modules/yarn/lib/cli.js:101464:78)
    at Object.findRc (/home/build/.nvm/versions/node/v10.24.1/lib/node_modules/yarn/lib/cli.js:101478:10)
    at getRcConfigForCwd (/home/build/.nvm/versions/node/v10.24.1/lib/node_modules/yarn/lib/cli.js:56957:74)
    at /home/build/.nvm/versions/node/v10.24.1/lib/node_modules/yarn/lib/cli.js:92827:56
    at Generator.next (<anonymous>)
    at step (/home/build/.nvm/versions/node/v10.24.1/lib/node_modules/yarn/lib/cli.js:310:30)
----------------------------------
Full logs are accessible in the console container at path :-
  stdout: /tmp/my127ws-stdout.txt
  stderr: /tmp/my127ws-stderr.txt
```

Introducing a `sleep 1` during these two commands lets it complete successfully.

I think we need to wait for the usermod/chown that happens in entrypoint.dynamic.sh to finish upon console container startup, before we continue onto the `app build` step.

marcomc:
Resolves #666 
Introduce Docker HealthCheck for local builds of the `console`

## The health check process:
1. The HealthCheck action steps generate a measurable condition during the execution of the Dynamic Entrypoint
2. The HealthCheck test run a command that verify if the health check condition is met
3. `docker-composer` will wait for the HealthCheck command to find the `console` container in a healthy status before releasing control over the containers to the following commands

## Defaults:

> All HealthCheck action, test and timing parameters are configurable (see `_src/base/CHANGELOG.md`)

* The HealthCheck action is set to create the `/HEALTHCHECK.txt` file
* The HealthCheck test command is set to launch the `/healthcheck.test.sh` script
* The HealthCheck `healthcheck.test.sh` is populated with steps that check if `/HEALTHCHECK.txt` has been created

 > This health check considers the `console` healthy when all the commands run inside the Dynamic Entrypoint have completed, with the creation of the `/HEALTHCHECK.txt` file as the last of the steps. If the file exists everything else has completed.
The HealthCheck can be much more complex if we need a more accurate health condition


src/_base/CHANGELOG.md
```diff
@@ -0,0 +1,15 @@
...
+
+* `attribute('console.healthcheck.enabled')` - Defaults to `true` only for local builds
+* `attribute('console.healthcheck.action.steps')` - Steps necessary to create a measurable condition to test later
+* `attribute('console.healthcheck.test.script')` - The path to the script that will host the health checks steps
+* `attribute('console.healthcheck.test.command')` - The command in array format that will run the healthcheck - Defaults to the test script path
+* `attribute('console.healthcheck.test.steps')` - Steps necessary to test the measurable condition to determine the health of the container
+* `attribute('console.healthcheck.test.start_period')` - Defaults to 5s
+* `attribute('console.healthcheck.test.interval')` - Defaults to 5s
+* `attribute('console.healthcheck.test.timeout')` - Defaults to 3s
+* `attribute('console.healthcheck.test.retries')` - Defaults to 10
```
andytson-inviqa, line 13: service.console.healthcheck for standard service use

src/_base/docker/image/console/root/healthcheck.test.sh.twig
```diff
@@ -0,0 +1,14 @@
...
+    # run any command required which result can be monitored by Docker's HEALTHCHECK
+    {% for step in @('console.healthcheck.test.steps') -%}
+    {{ step|raw }}
+    {% else -%}
+    echo "No healthcheck test steps declared"
+    {% endfor %}
+
+}
+
+healthcheck_test
```
andytson-inviqa: we already have https://github.com/inviqa/harness-base-php/blob/1.5.x/src/_base/docker/image/console/root/lib/task/state.sh, could we not just make that templateable instead?

src/_base/CHANGELOG.md
marcomc: I originally used that 
but give that `entrypoint` and `healthcheck` are Docker components at the same level
and that the `default` (empty) values for the `entrypoint` where set at the layer higher then `services` I thought it would be consistent to put the `healthcheck` at the same level as `entrypoint`

is that not appropriate?

src/_base/harness/attributes/common.yml
```diff
@@ -297,6 +297,21 @@ attributes.default:
       steps: []
 
   console:
+    healthcheck:
+      # we want that enabled only for local builds
+      enabled: false
+      action:
+        steps:
+          - echo 'Entrypoint tasks completed' > /tmp/HEALTHCHECK.txt
```
andytson-inviqa: do we definitely have use cases we can think of other than for end state indicators like this?

src/_base/docker/image/console/root/healthcheck.test.sh.twig
marcomc: State refers to the state of the `application` as an entity (in fact it checks the database in this implementation of `state.sh`)
I think that checking the `health` state of the container is another level then then the state of the application.

src/_base/CHANGELOG.md
andytson-inviqa: hmm I see `console...` already in use, so can see why it is as it is

src/_base/harness/attributes/common.yml
marcomc: This was the most basic check I could think of, as explained in the description of this PR
I am happy to implement a more precise and truthful test of health.
This is a basic example that works. just it.

andytson-inviqa: note this isn't the health check itself but the trigger that indicates its suitably started

src/_base/docker/image/console/root/healthcheck.test.sh.twig
andytson-inviqa: state refers the the health (state = healthy or not healthy) of the application. There's an arguable case for removing the db check as not really relevant to console for 99% of it's running time

src/_base/harness/attributes/common.yml
marcomc: yes the `action` is meant to generate a measurable state.
In this case we are considering `console` "healthy" when all the `usermod` and `groupmod` commands that are inside the dynamic entrypoint to be complete

Adding a function that is executed after those commands and that creates a `semaphore` file was the most accurae way to create a measurable think to say "oh, it's healthy"
of course, the `healthcheck_action` is customisable and also can be left doing nothing
and we simply can customise the `healthcheck_test` to check anything else we deem as a health parameter 

I see `healthcheck_action` as a hook where we can do something custom to generate a health factor IF we need to. and now we needed to (I couldn't think about anything smarter)

src/_base/docker/image/console/root/healthcheck.test.sh.twig
andytson-inviqa: i.e. console = sleep infinity, not the app

marcomc: exactly my point
`state` -> for test the health of the `application`
Docker `HEALTHCHECK` -> for test the health of the `console` container

with state `console` knows if the app is ok or not
but first we need to know if `console` is ok or not

I agree with you that the dabase connecttion is not a direct indicator of the health of the application, but I think we can address this separately

src/_base/harness/attributes/common.yml
andytson-inviqa: `I am happy to implement a more precise and truthful test of health.` yes but testing the health is the healthcheck's responsibility, rather than the entrypoint. Here is just a startup indicator, for a console which has no application running, so I can't see there need for anything other than indicating the process has completed.

andytson-inviqa: it's also running before console.entrypoint.steps, so isn't able to verify they have completed, just the dynamic bits like the chown (and no other user configurable steps other than the health check actions)

src/_base/docker/image/console/root/healthcheck.test.sh.twig
marcomc: > i.e. console = sleep infinity, not the app
I don't understand the point you make with this comment

andytson-inviqa: what's the application you're referring to? `sleep`?

marcomc: are you suggesting that we do not use `state` to test the application health and that we dedicate it to test the container health?
I think it might confuse people, I would prefer to leave `state` to test the health of the app, just update to a command that test other things rather then, or other then the database status

src/_base/harness/attributes/common.yml
marcomc: point taken!
what if I move `healthcheck_action()` to `entrypoint.sh.twig` and action it only if the environment is dynamic?

src/_base/docker/image/console/root/healthcheck.test.sh.twig
andytson-inviqa: webapp container is the application, cron container is the application etc.

src/_base/harness/attributes/common.yml
marcomc: > Here is just a startup indicator, for a console which has no application running,

yes

> so I can't see there need for anything other than indicating the process has completed.

true, for now thing more seems needed
I left it customisable just in case we might need it in the future

src/_base/docker/image/console/root/healthcheck.test.sh.twig
andytson-inviqa: they have their own checks they would use in their own containers which aren't the docker/image/console/... folder scripts (like a tcp/http probe for webapp but just for k8s)

src/_base/harness/attributes/common.yml
andytson-inviqa: I think the realm of the scripting of this would unlikely go beyond a startup indicator file, so could be hard coded in the entrypoint.sh.twig even, regardless of dynamic/static modes

src/_base/docker/image/console/root/healthcheck.test.sh.twig
marcomc: My worry is that we hook the Docker healhcheck to the execution of the `state.sh` script
and in any project, somebody that wants to run the the `ws`'s `state` command to check the health or the app also customise the `state.sh` script to have batter indicators of health.

Doing so it might compromise the health indicator of the container itself.
Keeping them separate avoids confusion
and adds flexibility

that is my rational 

If you tell that this is not a relevant issue
I can remove the `healthcheck.sh.twig` script and alter the `state.sh` to be a twig template and use that in Docker HEALTHCHECK as I think this is what you are suggesting.

src/_base/harness/attributes/common.yml
andytson-inviqa: and console.entrypoint.steps would then cater to anything that we can't think of yet.

marcomc: > and console.entrypoint.steps would then cater to anything that we can't think of yet.
you mean that `entrypoint.steps` could do what the `healthcheck.action.steps` are meant to do now?

if that case I agree that we can
1. hardcode the `healthcheck_action` function and steps in `entrypoint.sh.twig`
2. if a project needs a custom indicator of the `console` health rather than `/HEALTCHECK.txt` they can add those steps into `entrypoint.steps`
3. and modify `healthcheck.test.steps` or `healthcheck.test.command` to look for the custom indicator

**Let me know if my interpretation of your comment is correct, and I will implement it**

src/_base/docker/image/console/root/healthcheck.test.sh.twig
andytson-inviqa: https://github.com/search?p=1&q=org%3Ainviqa+task_state&type=Code we can see here the file hasn't been overriden anywhere yet for private projects.

I also think this startup indicator is as relevant to k8s as it is docker, so having the check in the app state command is fine

src/_base/harness/attributes/common.yml
andytson-inviqa: yes, though I think something that more simply describes the state as filename would be better than `healthcheck.txt`, e.g. `/run/started`

src/_base/docker/image/console/root/healthcheck.test.sh.twig
marcomc: moved the health check tests into `task_state' (or "app state")

src/_base/harness/attributes/common.yml
marcomc: moved the health check "checkpoint" into an embedded function in "entrypoint.sh.twig"

src/_base/harness/scripts/enable.sh.twig
```diff
@@ -85,7 +85,8 @@ dynamic_console()
 
     find .my127ws/docker/image/ -type d ! -perm u+rwx,go+rx,o-w -exec chmod u+rwx,go+rx,o-w {} +
 
-    passthru docker-compose up --build -d console
+    passthru docker-compose up --build -d console --wait
+    # docker-compose will wait for `console` to be in healty status if healthcheck is enabled
```
andytson-inviqa: comments should be before a line they are referring to, or otherwise in-line (but former is best here)
```suggestion
    # docker-compose will wait for `console` to be in healty status if healthcheck is enabled
    passthru docker-compose up --build -d console --wait
```

src/_base/docker/image/console/root/lib/task/state.sh.twig
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+function test_ok()
+{
+    # run any command required which result can be monitored by Docker's HEALTHCHECK
+    {% for step in @('console.state.steps') -%}
+    {{ step|raw }}
+    {% else -%}
+    echo "No state test steps declared"
```
andytson-inviqa:
```suggestion
    :
```
I don't see a point in this log line

src/_base/harness/attributes/environment/local.yml
```diff
@@ -3,4 +3,7 @@ attributes:
   app:
     build: dynamic
     mode: development
-    version: develop
\ No newline at end of file
+    version: develop
+  console:
+    healthcheck:
+      enabled: true
```
andytson-inviqa, line 9: can it be enabled for helm and CI as well? It's useful for seeing the deployment readiness, and CI should act the same as local

andytson-inviqa: technically helm way is enough being a common `console.healthcheck.enabled`, as it'll get the same file CI was built from

andytson-inviqa: note though that the helm way is hard coded to `app state`

andytson-inviqa:
I'm a bit nervous of the PR still, but now this sort of thing also needs to be a docker harness thing

PR #742, opened 2023-01-27
Switch over to docker compose v2

andytson-inviqa:
This expects docker compose v2 to be installed as a docker plugin rather than directly running it's binary.

### DIFF
--- a/src/_base/application/skeleton/README.md.twig
+++ b/src/_base/application/skeleton/README.md.twig
@@ -19,7 +19,7 @@ Please follow the below steps to get started, if you encounter any issues instal
 - A working Docker setup
   - On macOS, use [Docker Desktop for Mac](https://docs.docker.com/docker-for-mac/install/).
   - On Linux, add the official Docker repository described under the "Server" section on [Install Docker Engine](https://docs.docker.com/engine/install/) and install the "docker-ce" package.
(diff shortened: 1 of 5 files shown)


src/_base/application/skeleton/README.md.twig
```diff
@@ -19,7 +19,7 @@ Please follow the below steps to get started, if you encounter any issues instal
 - A working Docker setup
   - On macOS, use [Docker Desktop for Mac](https://docs.docker.com/docker-for-mac/install/).
   - On Linux, add the official Docker repository described under the "Server" section on [Install Docker Engine](https://docs.docker.com/engine/install/) and install the "docker-ce" package.
-    You will also need to have a recent [docker-compose](https://docs.docker.com/compose/install/) version - at least `1.26.0`.
+    You will also need to have a recent [docker compose](https://docs.docker.com/compose/install/) version - at least `1.26.0`.
```
kierenevans:
```suggestion
    You will also need to have a recent [docker compose](https://docs.docker.com/compose/install/) version - at least `v2.20.3`.
```

andytson-inviqa:
technically the harness-docker switch will do this anyway, but at least documents the change while harness-base-php should mention most docker changes itself prior to 2.0.1+

PR #744, opened 2023-02-22
Update Spryker demoshop to 202212.0-p2

andytson-inviqa:
may also need

- [x] To see if the config_local DB env vars are still being read correctly (from a report from @rocketage)


src/spryker/harness.yml
```diff
@@ -26,7 +26,7 @@ attributes:
   spryker:
     salt: 2tkqCCRKy5rT4wMVz8KTCh8r3sJGkL5v
     demoshop-url: https://github.com/spryker-shop/b2c-demo-shop.git
-    demoshop-version: "202211.0"
+    demoshop-version: "202212.0-p1"
```
andytson-inviqa:
```suggestion
    demoshop-version: "202212.0-p2"
```

src/spryker/application/skeleton/composer-harness.json.twig
```diff
@@ -3,6 +3,18 @@
...
+    "spryker/discount-promotions-rest-api": "^1.4.1",
+    "spryker/product-cart-connector": "^4.9.1",
+    "spryker/product-configuration-storage": "^1.0.2",
+    "spryker/gift-card": "^1.6.1",
+    "spryker/shipments-rest-api": "^1.6.1",
+    "spryker/payment": "^5.10.1",
+    "spryker/message-broker-aws": "^1.3.2",
+    "spryker/tax-product-connector": "^4.5.1",
+    "spryker/security-oauth-user": "^1.1.2",
+    "spryker-shop/storage-router": "^0.1.3",
```
andytson-inviqa: I don't think this applies to all spryker versions

andytson-inviqa: maybe no need to worry. The skeleton generally doesn't handle old versions anyway.

Akeneo and Drupal allow different versions

src/spryker/application/skeleton/src/Inviqa/Yves/EventDispatcher/EventDispatcherDependencyProvider.php
```diff
@@ -17,7 +17,7 @@ class EventDispatcherDependencyProvider extends PyzEventDispatcherDependencyProv
     /**
      * @override
      *
-     * @return \Spryker\Shared\EventDispatcherExtension\Dependency\Plugin\EventDispatcherPluginInterface[]
+     * @return array<\Spryker\Shared\EventDispatcherExtension\Dependency\Plugin\EventDispatcherPluginInterface>
```
andytson-inviqa, line 20: :( I hope coding standards doesn't force this `array<T>` instead of `T[]` now

andytson-inviqa: although technically it should be the namespace `use` as well, so namespace not needed in type

src/spryker/application/skeleton/composer-harness.json.twig
```diff
@@ -68,7 +78,7 @@
     "test-quality": [
       "@generate",
       "@phpstan",
-      "APPLICATION_ENV=development vendor/bin/console code:sniff:style",
+      "APPLICATION_ENV=development vendor/bin/console code:sniff:style src/Inviqa",
```
andytson-inviqa: Inviqa namespace would probably not be used by projects, but also Pyz can contain our own stuff too

andytson-inviqa: happy though an alternative that does all src/ but Pyz

src/spryker/application/skeleton/src/Inviqa/Yves/EventDispatcher/EventDispatcherDependencyProvider.php
g-foster2020: That was a modification due to code standards complaining, yes.

src/spryker/application/skeleton/composer-harness.json.twig
g-foster2020: Hopefully they can be removed on next demoshop update... or at least the patch versions up to spryker/tax-product-connector that fix the issues with dataBuilder using parentheses.

g-foster2020:
Rick notes that "Yves and Zed are serving responses, but Yves has no data in redis and Zed does not seem to have the correct CSS assets" from this testing. So we need someone (I will do it if time) to build from scratch and debug the remaining issues.

rocketage:
Sorry, another tiny point from me. In my `docker-compose.yml` the `php-fpm` service seems to be duplicated. I have a block on line 174 and a duplicate on line 550. This doesn't seem to cause too many problems, but things like `docker-compose exec php-fpm bash` do not work, but do work if the duplicate block is removed.

This issue must be also on the 1.4.1 harness because I see the problem on this PR plus the 1.4.1 harness.

I'm hoping it's an easy fix, but I cannot see anything supicious and not sure how to debug this myself.

andytson-inviqa:
we've had that php-fpm duplication for years, though I'd thought we'd fixed it.

I'd be surprised if it was the thing preventing bashing in. Also it's keys are merged with the other one by docker compose, so two versions with the same data should be the same result. That said newer docker compose versions may be acting differently to v1 which we previously required full use of

andytson-inviqa:
I've checked, and the docker compose v2.15.1 that's in Docker desktop isn't affected by duplicate php-fpms. I think this should be another bug ticket though

andytson-inviqa:
ah found it, 2.17.2 (latest version) errors:
```
docker-compose-darwin-aarch64 exec php-fpm bash
yaml: unmarshal errors:
  line 267: mapping key "php-fpm" already defined at line 130
```

added a issue for it https://github.com/inviqa/harness-base-php/issues/761

g-foster2020:
I am still checking the application works and fixing some bugs.

src/spryker/application/skeleton/config/install/docker.yml.twig
```diff
@@ -91,17 +91,24 @@ sections:
...
             command: 'vendor/bin/console router:cache:warm-up:backend-gateway'
+            stores: true
 
         twig-cache-warmup:
             command: "vendor/bin/console twig:cache:warmer"
             stores: true
 
+        translator-generate-cache:
+            command: "vendor/bin/console translator:generate-cache"
+            stores: true
```
andytson-inviqa: This didn't have `stores: true` before. Does it need to now? https://github.com/inviqa/harness-base-php/blob/1.4.x/src/spryker/application/skeleton/config/install/docker.yml.twig#L111-L112

andytson-inviqa: main way to check is to check the folder it writes to. If it has per store folders then it probably needs to be true. If not, then it may be running the same thing X*X times for X stores

g-foster2020: I believe it does according to https://github.com/spryker/translator/blob/master/src/Spryker/Zed/Translator/TranslatorConfig.php#L100

and in our test harness I have
`root@27f7128bfaa7:/app# ls ./src/Generated/Zed/Translator/codeBucketInviqa`

`root@27f7128bfaa7:/app# ls ./src/Generated/Zed/Translator/codeBucket/Inviqa/Pyz`

I assume that is our dummy store?

andytson-inviqa: no it'd look more like a locale. Pyz is a spryker class override method

rocketage: I wouldn't expect it to need multistore for this but if in doubt check to see what the original demoshop is doing: https://github.com/spryker-shop/b2c-demo-shop/blob/master/config/install/docker.yml#L40

In this case looks like it's ok without it.

g-foster2:
Next task: 

Remove the per-command env vars (DEVELOPMENT_CONSOLE_COMMANDS=1) that have been used to get the build working and identify where they are needed.

Refactor as per Andy's suggestion

- for devs always true (maybe an app.build == ‘dynamic’ service environment variable expression)
- for build (a buildArg maybe, or just an export ..="true" build step) true
- for static runtime false (maybe an app.build == ‘dynamic’ service environment variable expression for static being false)

g-foster2020:
After some experimentation I found the minimum need for ENABLE_DEVELOPMENT_CONSOLE_COMMANDS which is in dynamic build (for developers) and on three specific commands in composer.json (these are used in `composer generate` and in `composer test-quality` which is run on the static build at runtime).

Jenkins build passes.

The application itself still needs further smoke testing.

src/spryker/application/skeleton/composer-harness.json.twig
```diff
@@ -58,7 +76,7 @@
     ],
     "generate": [
       "APPLICATION_ENV=development vendor/bin/console transfer:generate",
-      "APPLICATION_ENV=development vendor/bin/console dev:ide:generate-auto-completion"
+      "APPLICATION_ENV=development DEVELOPMENT_CONSOLE_COMMANDS=1 vendor/bin/console dev:ide:generate-auto-completion"
```
andytson-inviqa, line 79:
```suggestion
      "APPLICATION_ENV=development vendor/bin/console dev:ide:generate-auto-completion"
```
shouldn't be needed as the variable should be service environment

src/spryker/application/skeleton/composer-harness.json.twig
```diff
@@ -67,8 +85,8 @@
     "test-quality": [
       "@generate",
       "@phpstan",
-      "APPLICATION_ENV=development vendor/bin/console code:sniff:style",
-      "APPLICATION_ENV=development vendor/bin/console code:sniff:architecture"
+      "find src/ -maxdepth 1 -mindepth 1 -path 'src/Pyz' -prune -o  -print | xargs -n1 env APPLICATION_ENV=development DEVELOPMENT_CONSOLE_COMMANDS=1 vendor/bin/console code:sniff:style",
+      "APPLICATION_ENV=development DEVELOPMENT_CONSOLE_COMMANDS=1 vendor/bin/console code:sniff:architecture"
```
andytson-inviqa, line 89: DEVELOPMENT_CONSOLE_COMMANDS=1   shouldn't be needed as the variable should be service environment

src/spryker/harness/attributes/docker.yml
```diff
@@ -3,6 +3,7 @@ attributes:
     php-base:
       environment:
         APPLICATION_ENV: = @('spryker.mode')
+        DEVELOPMENT_CONSOLE_COMMANDS: "= @('app.build') == 'dynamic' ? 1 : 0"
```
andytson-inviqa:
```suggestion
        DEVELOPMENT_CONSOLE_COMMANDS: "= @('app.build') == 'dynamic' ? '1' : '0'"
```
environment variables are always strings. Docker compose may error if integer used (also perhaps v2 only)

src/spryker/application/skeleton/composer-harness.json.twig
g-foster2020: We call composer test-quality on the static build so I was following the logic of APPLICATION_ENV=development being set here.

PR #745, opened 2023-02-22
Add skeleton config/bundles.php and fix Akeneo 4

andytson-inviqa:
First composer install sets up skeleton symfony structure, losing the original file.


src/akeneo/harness.yml
```diff
@@ -113,7 +115,7 @@ attributes:
     build:
       when: -f "package.json"
       steps:
-        - export NODE_OPTIONS=--openssl-legacy-provider
+        - "= version_compare(@('akeneo.major_version'), '6', '>=') ? 'export NODE_OPTIONS=--openssl-legacy-provider' : ''"
```
g-foster2020, line 118: Assume the 6 here is intentional.

g-foster2020 (approved):

src/akeneo/harness.yml
andytson-inviqa: yes, 5 didn't need it, and was failing because of it

PR #747, opened 2023-03-13
Support ssh private key Jenkins credentials for chart publish

andytson-inviqa:


andytson-inviqa:
just realised this has a bug where if no publish environment variables, then the credential wont be added.

I hate having to wrap all conditions duplicated in the outer if, esp as a 3rd's going to be needed, so will have a think about how to do this without that

kierenevans (dismissed):

kierenevans (approved):

PR #751, opened 2023-03-20
Remove startup_core Drupal module and profile

andytson-inviqa:
This shouldn't be here


andytson-inviqa:
seems startup still used on projects for the more obvious name rather than the non-obvious one, so not all this needs removing

PR #752, opened 2023-03-21
Switch db.platform_version default if db.platform changed

andytson-inviqa:
Spryker was using postgres with PHP seeing the version as 8


kierenevans (approved):

andytson-inviqa:
naturally is blocked by the current spryker build bug

PR #755, opened 2023-03-21
Use tagged version of busybox

kierenevans:
Avoid using `:latest` as it's unpredictable.

Less important due to only using `chown` from busybox.

### DIFF
--- a/src/_base/helm/app/templates/application/cron/deployment.yaml
+++ b/src/_base/helm/app/templates/application/cron/deployment.yaml
@@ -29,7 +29,7 @@ spec:
       {{- if not (eq "" (include "application.volumes.wwwDataPaths" $)) }}
       initContainers:
       - name: cron-volume-permissions
-        image: busybox:1.36
+        image: {{ .initContainers.volumePermissions.image }}
         command:
         - "/bin/chown"
         - "-R"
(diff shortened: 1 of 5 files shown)


src/_base/helm/app/templates/application/cron/deployment.yaml
```diff
@@ -29,7 +29,7 @@ spec:
       {{- if not (eq "" (include "application.volumes.wwwDataPaths" $)) }}
       initContainers:
       - name: cron-volume-permissions
-        image: busybox
+        image: busybox:1.36
```
andytson-inviqa: should be configurable if not latest

andytson-inviqa: 1.35/1.36 are marked "unstable" as well

kierenevans: Will sort :+1:

kierenevans:
Build failed, will take a look

andytson-inviqa (approved):

src/_base/helm/app/templates/application/cron/deployment.yaml
andytson-inviqa: erm, sorry as mentioned 1.35 also "unstable". 1.34 seems the most recent "stable"

kierenevans: 1.35 is the earliest supported tag on https://hub.docker.com/_/busybox

PR #759, opened 2023-03-23
Fix mutagen.sh to work with mutagen version 0.15.0 up

### LINKED ISSUE #757: mutagen.sh no longer cleans up duplicate sessions when mutagen version >= 0.15.0
After version 0.15.0 of mutagen, labels are only displayed if the `-l` flag is used.

https://github.com/mutagen-io/mutagen/compare/v0.14.0...v0.15.0#diff-ff22205d1c8f5db2ec359d969fff6a01b8f42e44cbf006f05cbe345841d94606L69

Currently mutagen.sh relies on the labels being displayed in order to detect and offer to clean up duplicate sessions.

https://github.com/inviqa/harness-base-php/blob/1.5.x/src/_base/harness/scripts/mutagen.sh#L93

`... grep -B4 "URL: $(pwd)" | grep io.mutagen.project ...`

`io.mutagen.project` is only displayed if `-l` flag is used on later versions of mutagen and it isn't in the four lines before URL because a lot of other data is displayed too.

---

Similar issues with the forward list command.

g-foster2020:
Fixes #757

### DIFF
--- a/src/_base/harness/scripts/mutagen.sh
+++ b/src/_base/harness/scripts/mutagen.sh
@@ -86,11 +86,11 @@ clean_existing_projects()
     local SYNC_LIST
     for SYNC_NAME in "${SYNC_NAMES[@]}"; do
         # List syncs based on name
-        SYNC_LIST="$(mutagen sync list "$SYNC_NAME" 2> /dev/null || true)"
+        SYNC_LIST="$(mutagen sync list -l "$SYNC_NAME" 2> /dev/null || true)"
         # Check if there are entries left
         if [ "$(echo "$SYNC_LIST" | grep --count "URL: $(pwd)" | awk '{ print $1 }')" -gt 0 ]; then
             # Build an array of sync session IDs to clean up
-            while IFS='' read -r line; do EXISTING_PROJECT_LABELS+=("$line"); done < <(echo "$SYNC_LIST" | grep -B4 "URL: $(pwd)" | grep io.mutagen.project | awk '{print $1"="$2}' | sed s/:=/=/)
+            while IFS='' read -r line; do EXISTING_PROJECT_LABELS+=("$line"); done < <(echo "$SYNC_LIST" | grep -A6 "Name: $SYNC_NAME" | grep io.mutagen.project | awk '{print $1"="$2}' | sed s/:=/=/)
         fi
     done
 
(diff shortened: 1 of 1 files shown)


kierenevans:
I was wondering if the `-l` option existed in v0.11.8, and it does: https://github.com/mutagen-io/mutagen/blob/v0.11.8/cmd/mutagen/sync/list.go#L215

src/_base/harness/scripts/mutagen.sh
```diff
@@ -86,11 +86,11 @@ clean_existing_projects()
     local SYNC_LIST
     for SYNC_NAME in "${SYNC_NAMES[@]}"; do
         # List syncs based on name
-        SYNC_LIST="$(mutagen sync list "$SYNC_NAME" 2> /dev/null || true)"
+        SYNC_LIST="$(mutagen sync list -l "$SYNC_NAME" 2> /dev/null || true)"
         # Check if there are entries left
         if [ "$(echo "$SYNC_LIST" | grep --count "URL: $(pwd)" | awk '{ print $1 }')" -gt 0 ]; then
             # Build an array of sync session IDs to clean up
-            while IFS='' read -r line; do EXISTING_PROJECT_LABELS+=("$line"); done < <(echo "$SYNC_LIST" | grep -B4 "URL: $(pwd)" | grep io.mutagen.project | awk '{print $1"="$2}' | sed s/:=/=/)
+            while IFS='' read -r line; do EXISTING_PROJECT_LABELS+=("$line"); done < <(echo "$SYNC_LIST" | grep -A6 "Name: $SYNC_NAME" | grep io.mutagen.project | awk '{print $1"="$2}' | sed s/:=/=/)
```
kierenevans: Does this run the risk of cleaning up `test2` when the sync name is `test`?

I guess that was true of the pwd approach too, but it's hopefully less likely to have a collision on path there

g-foster2020:
> I was wondering if the `-l` option existed in v0.11.8, and it does: https://github.com/mutagen-io/mutagen/blob/v0.11.8/cmd/mutagen/sync/list.go#L215

That's good. I wasn't expecting backwards compatibility. I thought about trying to compare the semver but gave up on that.

src/_base/harness/scripts/mutagen.sh
g-foster2020: Yes :( need to update the regexp

kierenevans (approved):

PR #763, opened 2023-03-30
Remove duplicate php-fpm container

### LINKED ISSUE #761: docker-compose.yml fails to parse in compose v2.17.0+
Since compose 2.17.0, the duplicate php-fpm services in docker-compose.yml result in an error:
```
yaml: unmarshal errors:
  line 267: mapping key "php-fpm" already defined at line 130
```

Previous versions of compose were fine.

This is as webapp and php-fpm are both enabled, but webapp is what was supposed to originally include/exclude php-fpm service

kierenevans:
Fixes #761 

services.php-fpm.enabled is true for helm chart purposes since #609, so will be included as part of the main application.yml.twig loop over enabled services.

A few alternatives:
1. Set enabled to false in services for docker and true in helm for pipeline.base.services (potential breaking change?)
2. Add an if statement in the application.yml.twig to skip over php-fpm.

### DIFF
--- a/src/_base/_twig/docker-compose.yml/service/webapp.yml.twig
+++ b/src/_base/_twig/docker-compose.yml/service/webapp.yml.twig
@@ -1,3 +1,2 @@
 {% set blocks  = '_twig/docker-compose.yml/' %}
 {% include blocks ~ 'service/nginx.yml.twig' %}
-{% include blocks ~ 'service/php-fpm.yml.twig' %}


src/_base/_twig/docker-compose.yml/service/webapp.yml.twig
```diff
@@ -1,3 +1,2 @@
 {% set blocks  = '_twig/docker-compose.yml/' %}
 {% include blocks ~ 'service/nginx.yml.twig' %}
-{% include blocks ~ 'service/php-fpm.yml.twig' %}
```
andytson-inviqa, line 3: I think its a bit confusing only having nginx here, and hence a similar problem if we add enabled to nginx.

andytson-inviqa:
or 3. set a variable in webapp.yml which is checked to conditionally render php/nginx if webapp service is enabled and the variable is set

andytson-inviqa:
I'll try a PR with that to compare

andytson-inviqa:
went with #765 instead

PR #764, opened 2023-03-30
Upgrade to NGINX 1.23

kierenevans:
https://hub.docker.com/_/nginx no longer lists 1.21 as a supported tag.


andytson-inviqa (approved):

andytson-inviqa:
```
 *) Change: now nginx combines arbitrary header lines with identical
       names when sending to FastCGI, SCGI, and uwsgi backends, in the
       $r->header_in() method of the ngx_http_perl_module, and during lookup
       of the "$http_...", "$sent_http_...", "$sent_trailer_...",
```

Interesting, might catch out bad php http header implementations

PR #767, opened 2023-04-03
Fix support for compose v2 in drupal test-unit command

andytson-inviqa:


andytson-inviqa:
tested with compose v2 in #742

andytson-inviqa:
it's disappointing how complex this test-unit command is. Ideally it should be simple to allow running fast in repetition.

kierenevans:
It's related to the usage of simpletest, which requires files generated by simpletest (via the phpunit command) in docroot/sites/simpletest/ to be available in the php-fpm container.

That plus the lack of development dependencies having been installed into the php-fpm container.

In the development environment this is a mountpoint shared between both, but in the pipeline it's more tricky.

Perhaps we simply share a volume between them in the pipeline instead?

kierenevans (approved):

PR #768, opened 2023-04-17
Delete console container after quality tests

g-foster2020:
Ensure console container is rebuilt after having dev dependencies installed for quality tests.

### DIFF
--- a/test
+++ b/test
@@ -100,6 +100,7 @@ function test_harness_acceptance()
     if ! [[ "$harness" =~ ^(magento1)$ ]]; then
       install_console
       run_project_quality_tests "$harness"
+      destroy_environment
     fi
     install_environment
     check_environment_started "$harness" "$mode"
@@ -219,6 +220,12 @@ function restart_environment()
     cd "${path_test}"
     ws disable
     ws enable
+)
+
+function destroy_environment()
+(
+    cd "${path_test}"
+    ws destroy
 )
 
 function setup_dynamic_mountpoint()
--- a/test
+++ b/test
@@ -100,7 +100,7 @@ function test_harness_acceptance()
     if ! [[ "$harness" =~ ^(magento1)$ ]]; then
       install_console
       run_project_quality_tests "$harness"
-      destroy_environment
+      remove_console
     fi
     install_environment
     check_environment_started "$harness" "$mode"
@@ -142,6 +142,12 @@ function install_console()
 (
     cd "${path_test}"
     ws enable console
+)
+
+function remove_console()
+(
+    cd "${path_test}"
+    docker-compose rm --stop console
 )
 
 function install_environment()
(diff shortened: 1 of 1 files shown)


test
```diff
@@ -221,6 +222,12 @@ function restart_environment()
     ws enable
 )
 
+function destroy_environment()
```
andytson-inviqa: can we go for something a lot less impactful, like
```
docker-compose rm --stop console
docker-compose up -d console
```

note tho $COMPOSE_BIN

andytson-inviqa: that way the console container doesn't need rebuilding or in mutagen's case a full resync of the volume

andytson-inviqa: however... acceptance tests also need the dev dependencies, so I'm not sure this is the answer

andytson-inviqa (changes requested):
destroy is too destructive and acceptance tests need dev dependencies

g-foster2020:
> destroy is too destructive and acceptance tests need dev dependencies

I can't see how the acceptance tests can require the dev dependencies because this works?

andytson-inviqa:
there's something more to it about these changes then, as obviously `behat` and co are dev dependencies and so would exist in console container

test
```diff
@@ -143,6 +144,12 @@ function install_console()
     ws enable console
 )
 
+function remove_console()
+(
+    cd "${path_test}"
+    docker-compose rm --stop console
```
andytson-inviqa, line 150: this is missing `docker-compose up -d console` from my previous suggestion, so would probably force a rebuild of console still?

andytson-inviqa:
you're missing a harness label so no harnesses were tested

g-foster2020:
Yes I can see run_project_quality_tests gets run twice. Before ws install and then during the acceptance tests. The issue was having it run before without cleaning up.

g-foster2020:
> Yes I can see run_project_quality_tests gets run twice. Before ws install and then during the acceptance tests. The issue was having it run before without cleaning up.

sorry ignore this ... run for magento only during acceptance tests

g-foster2020:
This does seem to be the wrong solution.

The issue is that ws install didn't fail when it should have done, because the quality tests were run first and they installed dev dependencies which hid an issue.

I think it would be better to either remove the dev dependencies after they are needed or have a clean environment before running install_environment.

test
andytson-inviqa: i.e. you only want to clear the changes to the console container based on what you're doing, so a rm and an up gets it to that state (and container is running again after, rather than potentially tripping on on a enable.sh check)

andytson-inviqa:
Also the other thing is that project Jenkinsfile also has the dev dependencies installed behaviour. The test.sh is recreating those Jenkinsfiles steps in bash

g-foster2020:
It would be clearer to revert these changes and have a specific test_harness_install step.

g-foster2020:
This was a misguided solution due to lack of experience with the test script. PR will be closed and an issue opened for future consideration.

PR #770, opened 2023-04-26
Update compose ps command to show container hash for DC v2.14

rocketage:
## Background

It seems that the behaviour of `docker-compose ps --quiet` has changed from v2.13 to v2.14 of `docker-compose`

In v1 this command would return the stopped containers, but in v2.14 it only returns the containers if they are running.

The fix is to use the `--all` flag. With that the command lists all stopped containers.

## Changes
1. Add `--all` flag to all `docker-compose ps` commands in `enable.sh`

### DIFF
--- a/src/_base/harness/scripts/enable.sh.twig
+++ b/src/_base/harness/scripts/enable.sh.twig
@@ -8,7 +8,7 @@ console_enabled()
     if [ "$APP_BUILD" = dynamic ]; then
         [ -f .my127ws/.flag-console-built ]
     else
-        [ -n "$("${COMPOSE_BIN[@]}" ps --quiet console)" ]
+        [ -n "$("${COMPOSE_BIN[@]}" ps --quiet --all console)" ]
     fi
 }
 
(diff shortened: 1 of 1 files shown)


rocketage:
I was able to avoid the first build error by updating `compose.json` but the next build error is this from the Spryker test:

```
Exception - Exception: Value mismatch for "Address.firstName" transfer property. Value1: "firstName()"; Value2: "firstName". To fix this, search for 'property name="firstName"' in the code base and fix the wrong one. 

in /app/vendor/spryker/transfer/src/Spryker/Zed/Transfer/Business/Model/Generator/TransferDefinitionMerger.php (160)
```

I assume we need to merge this first to get a green build: https://github.com/inviqa/harness-base-php/pull/744

g-foster2020:
> 
> I assume we need to merge this first to get a green build: #744

Yes unless Spryker tests are skipped.

andytson-inviqa (approved):

src/_base/harness/scripts/enable.sh.twig
```diff
@@ -77,7 +77,7 @@ dynamic_console()
 {
     if console_enabled; then
         # ensure it is started
-        if [ -z "$("${COMPOSE_BIN[@]}" ps --quiet console)" ]; then
+        if [ -z "$("${COMPOSE_BIN[@]}" ps --quiet --all console)" ]; then
```
andytson-inviqa, line 80: erm, not actually relevant here as its not started if it's not started, but as it's only reproducing same behaviour as before, there's no new bug

PR #771, opened 2023-04-26
Remove phpcsutils explicit dependency now that implicit dependency works

andytson-inviqa:
php-compatibility required previously an alpha package, which we added this explicit dependency to solve, however now that version is not supported by php-compatibility

In hindsight, `~1.0.0@alpha, ^1.0` might have been better

### DIFF
--- a/src/_base/harness/attributes/common.yml
+++ b/src/_base/harness/attributes/common.yml
@@ -235,7 +235,7 @@ attributes.default:
 
   php:
     version: "8.0"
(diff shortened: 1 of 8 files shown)


g-foster2020 (dismissed):

g-foster2020 (dismissed):

g-foster2020 (approved):

andytson-inviqa:
going to split this #772 out for a passing build, to avoid building everything multiple times

andytson-inviqa:
oh, forgot I'd gotten auto-merge enabled. oh well, will close #772 then

PR #772, opened 2023-04-28
Upgrade PHP 7.3 debian distros to buster

andytson-inviqa:
PHP 7.3 builds currently use stretch, but is EOL

We have buster images for them, so switch to buster


andytson-inviqa:
Also worth noting that my127/akeneo:7.3-fpm-stretch hasn't been updated in 3 years due to it not having a https://github.com/my127/docker-akeneo build, but was being used in this.

andytson-inviqa:
was merged in #771

PR #776, opened 2023-05-16
Add read committed isolation level for drupal 10

kierenevans:
New recommendations for drupal 10

### DIFF
--- a/src/drupal/application/overlay/docroot/sites/default/settings.local.php.twig
+++ b/src/drupal/application/overlay/docroot/sites/default/settings.local.php.twig
@@ -15,6 +15,11 @@ $databases['default']['default'] = array (
   'port'      => '3306',
   'namespace' => 'Drupal\\Core\\Database\\Driver\\mysql',
   'driver'    => 'mysql',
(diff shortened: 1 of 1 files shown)


src/drupal/application/overlay/docroot/sites/default/settings.local.php.twig
```diff
@@ -15,6 +15,11 @@ $databases['default']['default'] = array (
   'port'      => '3306',
   'namespace' => 'Drupal\\Core\\Database\\Driver\\mysql',
   'driver'    => 'mysql',
+{% if version_compare(@('drupal.major_version'), '10', '>=') %}
+  'init_commands' => [
+    'isolation_level' => 'SET SESSION TRANSACTION ISOLATION LEVEL READ COMMITTED',
+  ],
+{% endif %}
```
andytson-inviqa, line 22: note also this might get wiped unless `$databases['default']['default'] = ...` is removed from settings.php, as it's defined after settings.local.php is included

andytson-inviqa (approved):

PR #781, opened 2023-08-07
Replace php intermediate image build with multi-stage Dockerfile

andytson-inviqa:
So to remove most logic that docker builds can do internally more efficiently.

For Spryker's jenkins-runner this means a change from using console as base image to the app base (which was previously known as php-fpm)

This means fewer overrides of the docker harness will be needed in a subsequent PR.


marcomc (dismissed):
This looks clean and clear to me.

PR #785, opened 2023-08-18
Ensure all startup commands convert sidekick run to passthru

andytson-inviqa:
The /tmp/my127ws-std* files created during startup, when on failure, are not available.

Also this is a workaround to the bug in k8s 1.25 app-init/app-migrate not being able to write to these files when root user


kierenevans (approved):

andytson-inviqa:
seems to work correctly, drupal9's `drush cache:rebuild`:
* `docker compose run console app migrate` - passthrus
* `ws exec app migrate` - runs 
* `docker compose run console bash -c 'sleep infinity'`, then `docker exec -it {container name} app migrate` - runs 
  * this one just to test although SIDEKICK_VERBOSE exported to the init process and sub-processes, exec'ing doesn't get that variable. The default CMD's sleep infinity doesn't get the export as it doesn't need it

PR #786, opened 2023-08-18
Ensure static builds convert sidekick run to passthru

andytson-inviqa:
So that the logs are available in console on failure, rather than lost, and not embedded as files in the resulting image


kierenevans (dismissed):

andytson-inviqa:
still is writing to file

andytson-inviqa:
turns out it is working, just the ./test uses exec rather than run, and exec intentionally doesn't have `SIDEKICK_VERBOSE` set, so the files I saw were on the container not the image

PR #787, opened 2023-08-21
Simplify release process using Github released archives

andytson-inviqa:
Still requires versions updating in files, but doesn't require sub-tree forks

In the least this will be useful for testing pre-releases.

### DIFF
--- a/.github/workflows/release-archives.yml
+++ b/.github/workflows/release-archives.yml
@@ -1,0 +1,31 @@
+on:
+  release:
+    types:
+      - created
+
+name: Attach archives to release
+
+jobs:
+  build:
+    name: Build and upload archives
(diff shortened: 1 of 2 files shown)


andytson-inviqa:
~another bonus of the `./build --archive` is you can just reference the tgz file path locally as a harness/harnessLayer as well for testing locally~

actually no ws doesn't like any local file format path :(

g-foster2020 (dismissed):

andytson-inviqa:
hmm though it's got a similar directory structure (though no version number), `ws harness download` doesn't remove the first path, so e.g. results in `.my127ws/harness-drupal/*`

andytson-inviqa:
fixed and tested using https://github.com/my127/workspace/pull/162

PR #793, opened 2023-10-19
Remove redis save option so redis 7 can be used

andytson-inviqa:
There is a difference in config format for redis 7, and since save are it's defaults, it doesn't need supplying.

So remove the need for a workaround to supply save options if using redis 7


andytson-inviqa:
it seems the array options of save were unable to be unset without malformating, so this is necessary other than by overriding the redis service templates to hard code/fix the issue there

PeterThorndycraft (approved):

PR #798, opened 2023-12-08
Update PHP/Symfony/Wordpress PHP version to 8.2

andytson-inviqa:
and ensure compatibility for all harnesses for their PHP version

### DIFF
--- a/src/magento2/harness.yml
+++ b/src/magento2/harness.yml
@@ -45,7 +45,7 @@ attributes:
       - "= @('services.tideways.enabled') ? 'tideways' : ''"
       - redis
       - sockets
-    version: 7.4
+    version: '7.4'
   composer:
     auth:
       basic:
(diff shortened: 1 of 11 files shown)


src/magento2/harness.yml
```diff
@@ -45,7 +45,7 @@ attributes:
       - "= @('services.tideways.enabled') ? 'tideways' : ''"
       - redis
       - sockets
-    version: 7.4
+    version: '7.4'
```
kierenevans, line 48:
```suggestion
    version: '8.1'
```
(unless unsupported with our magento2 version - might need updating)

andytson-inviqa: will worry about that in another PR

andytson-inviqa: 8.2 instead https://github.com/inviqa/harness-base-php/pull/800

PR #803, opened 2024-01-26
feat(all): Add corepack support

andytson-inviqa:
nodejs is generally now being distributed with corepack, which takes over yarn and pnpm binaries and allows support for yarn 2+ without needing it to be committed to repositories, instead a "packageManager" package.json field will define the version used, and downloaded on execution of yarn or pnpm.

By default yarn 1 will be used, the same as yarn npm package. Currently, like node docker image, `corepack enable` needs to be run before package installation (or `corepack yarn ...`) to get this behaviour.

### DIFF
--- a/src/_base/docker/image/console/Dockerfile.twig
+++ b/src/_base/docker/image/console/Dockerfile.twig
@@ -52,7 +52,8 @@ RUN . {{ @('app.code_owner_home')}}/.nvm/nvm.sh \
  && nvm install {{ @('node.version') }} \
  && nvm use {{ @('node.version') }} \
  && nvm alias default {{ @('node.version') }} \
- && npm install -g yarn
+ && npm install -g corepack \
+ && corepack enable
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


src/_base/docker/image/console/Dockerfile.twig
```diff
@@ -52,7 +52,8 @@ RUN . {{ @('app.code_owner_home')}}/.nvm/nvm.sh \
  && nvm install {{ @('node.version') }} \
  && nvm use {{ @('node.version') }} \
  && nvm alias default {{ @('node.version') }} \
- && npm install -g yarn
+ && npm install -g corepack yarn \
+ && corepack pack --all
```
andytson-inviqa: consideration, pack only a detected used package manager only instead?

andytson-inviqa: perhaps pack not needed, as static build is including it in image on install

kierenevans (approved):

kierenevans (approved):

PR #805, opened 2024-02-22
fix(akeneo): add a temporary restriction on symfony/console version

andytson-inviqa:
symfony/console versions 5.4.35, 6.3.12, 6.4.3, 7.0.3 have a regression in loose InputOption shortcut typechecks, which misses `false` as no option.

Some Akeneo commands use false (incorrectly) instead of null for shortcut

### DIFF
--- a/src/akeneo/application/skeleton/composer.json.twig
+++ b/src/akeneo/application/skeleton/composer.json.twig
@@ -38,7 +38,8 @@
 {%- endif %}
   "require": {
     "php": ">= {{ @('php.version') }}",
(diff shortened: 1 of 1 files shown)


andytson-inviqa:
may not merge this as it could be fixed any time soon, as the fix is in the branches

kierenevans (dismissed):

andytson-inviqa:
no longer needed, symfony/console 7.0.4, 6.4.4, 5.4.36 has the fix. Oddly they didn't release 6.3 fix, but akeneo 5.4

PR #806, opened 2024-02-23
feat(spryker): Remove spryker harness

andytson-inviqa:
The harness has fallen into disrepair and needs an upgrade to fix, however is not used by Inviqa anymore, so instead development will move to https://github.com/inviqa/harness-spryker


kierenevans (dismissed):

andytson-inviqa:
Just a rebase conflict fix to remove the changed spryker files

kierenevans (approved):

PR #811, opened 2024-07-17
Backport v2 name addition and version removal from docker-compose.yml

andytson-inviqa:
PHP harnesses v2.0.0 will have this change (from using docker harness as parent layer), but this is needed earlier to get rid of annoying warnings


rgpjones (approved):
Looks great!

